PR #2477, opened 2015-01-05
Add gatling scale out .sh script and update doc

Nimrod007:


pdalpra:
There is a small issue with your PR : 

Linking the script properly from the documentation may prove a bit difficult, and does not work in its current state.

Could you please, for the moment, inline the script in the "Scaling Out" page ?
We'll sort it out after ;)

Could you also rebase your commits after that ?

Nimrod007:
add inline script + rebase, you can change later on and remove the inline script/ add link
thanks

pdalpra:
Thanks !

PR #2480, opened 2015-01-05
Core: Split Controller and Throttler responsibilities

slandelle:


slandelle:
@pdalpra Please review

PR #2485, opened 2015-01-07
Share some scopt-related code between App and Recorder

pdalpra:


slandelle:
LGTM, please rebase and merge

PR #2494, opened 2015-01-08
Setup Docker

DaveBlooman:
Setup Docker for users to create graphite graphs

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,22 @@
+Gatling Graphite Charts
+=========
+
+This is a generic metrics and logging tool for Load testing.  It uses [Graphite](http://graphite.readthedocs.org/en/latest/) and [Grafana](http://grafana.org/) for the front end, with [Carbon](http://graphite.wikidot.com/carbon) as the backend.  Nginx is used as a reverse proxy and is in it's own container.  [Elasticsearch](http://www.elasticsearch.org/) is used as the JSON store for the Grafana dashboards, this will mean that you don't need to restart the service when you want to add new data and dashboards.
+
+To build graphite-charts:
+
+```
+./build
+```
+
+To start:
+
+```
+./start
+```
+
+To stop:
+
+```
+./stop
+```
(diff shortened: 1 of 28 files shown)


pdalpra:
Hi,

Sorry it took so long, I finally had time to have a look at your PR.
The thing that bothers me most is the use of ElasticSearch in this context. Can't we simply go with a few JSON files instead of require a whole other container for storing the dashboards ?

DaveBlooman:
I agree.  I suppose the best thing to do is decide what data we want on the dashboard, or at least, what the simplest default setup is, then apply that and save the JSON as the default dashboard.  If people want to save dashboards, they can always export and upload, but for personal use, ES is probably overkill.

What should be the default setup for the dashboard?

pdalpra:
And, BTW, nothing would block them from modifying the container to edit the config to point to an ES cluster, should they want to ;)

I guess that displaying : 

* The number of active users
* A few global percentiles (95th, 99th and possibly 50th)

should suffice.

I was tempted to rely on `fig` for the container orchestration, but I guess this can wait until Docker Compose makes its way through to Docker itself...

DaveBlooman:
I stayed away from fig as it requires python right now.  I will create some generic dashboards

DaveBlooman:
OK, just playing around, perhaps something like this

screenshot 2015-01-21 16 40 12

slandelle:
@DaveBlooman It looks like you're using a wildcard for the request name, so Graphana fetches all of them. But then, what if a user has 50 different request names or more. I think we should stick to allRequests in the default set up.

IMHO, we should have:
* users (with per scenario detail)
* response times for all requests (min, max, 95 and 99 percentiles)
* number of responses per second for all requests (OK, KO, total)

WDYT?

DaveBlooman:
Sorry it's been a while, we have moved away from Graphite towards InfluxDB.  Just published a post on on a local setup, http://www.dblooman.com/docker/2015/02/22/real-time-gatling-results-with-docker-and-influxdb/ with associated Dockerfiles and setup https://github.com/DaveBlooman/gatling-docker.

slandelle:
Thanks David! We'll get this altogether as soon as we can!

PR #2498, opened 2015-01-09
Replace disableUrlEncoding with disableUrlEscaping, close #2497.

### LINKED ISSUE #2497: Doc: non-existing method disableUrlEncoding
The document says that we can use ``.disableUrlEncoding``,
> If you know that your urls are already properly encoded, you can disable this feature with ``.disableUrlEncoding``.
http://gatling.io/docs/2.1.2/http/http_protocol.html

but we can't compile the following file:
```scala
import io.gatling.http.Predef._

class BasicSimulation extends Simulation {
  val httpConf = http.baseURL("http://example.com").disableUrlEncoding
  val scn = scenario("Scenario Name").exec(http("request_1").get("/"))
  setUp(scn.inject(atOnceUsers(1)).protocols(httpConf))
}
```

Here is the error message:
> value disableUrlEncoding is not a member of io.gatling.http.config.HttpProtocolBuilder

mnogu:


slandelle:
Thanks!

PR #2513, opened 2015-01-15
Upgrade logback.xml, more verbose response loggers by default close #2509

### LINKED ISSUE #2509: HTTP: Improve logback.xml so response bodies are not discarded on debug
Currently, you actually have to enable DEBUG on `io.gatling.http.response`.

convenient:
Added commented out `DEBUG` / `TRACE` `io.gatling.http.response` loggers to the default `logback.xml`


slandelle:
Let's go this way for now, but it would be nice to have a single trigger. Will try to think of something.

Thanks!

convenient:
Aye t'would be cool to have a single trigger. But at least now people won't fall into the same trap that I did and spend hours trying to debug odd responses :+1:

PR #2546, opened 2015-02-04
SSE close action needs to wait a last event before being executed

ctranxuan:
Hello,
When using the SSE protocol, the ```close()``` action has to wait another event sent by the SSE server before being triggered. In other words, as long as the sse server has not sent another event, the close action is not executed.

For instance, if you take the example below:  

```scala
class ProxyStockmarketSimulation extends Simulation {
  val httpConf = http
                 .baseURL("http://blablabl.com")
                 .doNotTrackHeader("1")

  val scn = scenario(this.getClass.getSimpleName)
            .exec(sse("sse").open("/app/stockmarket/prices"))
            .exec(sse("close").close())

  setUp(scn.inject(rampUsers(1) over 1)).protocols(httpConf)
}
```
The SSE server will sent one event when ```open()``` is called but the ```close()``` will not been triggered as long as the SSE server has not sent another event.  Which is not the expected behaviour.

The pull request fixes the issue.

Best regards,
Cédric

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/action/sse/SseActor.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/action/sse/SseActor.scala
@@ -222,6 +222,9 @@ class SseActor(sseName: String) extends BaseActor with DataWriterClient {
           .applyUpdates(session)
           .copy(requestName = requestName, start = nowMillis, next = next)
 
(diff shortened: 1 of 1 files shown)


slandelle:
Hello @ctranxuan 

I fixed it in another way, see #2556. Thanks for reporting and proposing a fix!

ctranxuan:
Great! Thanks!

Issue #2591, opened 2015-03-02
HTTP: SSE blocking check does not stop after timeout


ctranxuan:
Hello,
With the following example:

```scala
package org.tests.gatling.sse

import io.gatling.core.Predef._
import io.gatling.http.Predef._

import scala.concurrent.duration._

class NewProxySimulation extends Simulation {
  val httpConf = http
  .baseURL("http://my-sse-server.com")

  val scn = scenario(this.getClass.getSimpleName)
  .exec(sse("sse").open("/app/stockmarket/prices")
                .check(wsAwait.within(10).until(1).regex("""event(.*)"""))
    )
           .exec(sse("close").close())
  setUp(scn.inject(rampUsers(1) over 1)).protocols(httpConf)
```
the simulation does not timeout as long as the SSE server does not send at least one SSE event.

slandelle:
Hi @ctranxuan 

What's your SSE implementation? It seems to me that the server should reply ASAP with the HTTP status code instead of waiting for the first message.

Do you have a reproducer?

slandelle:
Some information for the world renown expert @jfarcand, such a behavior might crash some Firefox. Atmosphere sends a white space payload in such a situation.

@ctranxuan Could you share a reproducer, please?

jfarcand:
@ctranxuan Which server is acting like that BTW?

ctranxuan:
Apologizes, it might be an issue in our side: the SSE message sent was messy. I need to confirm with our QA that the message is the origin of our issue or not.
But at first sight, we thought it came from the Gatling SSEHandler. In the meantime, I had some issue in using a patched version of Gatling (some maven resolution stuff, I assume), so I come back to you once I get the QA feedback (it might take a week: I am not in our office for this period of time).

ctranxuan:
Hello,
We check with our QA: it was indeed an issue with the SSE streams that was not well formatted.
The issue can be closed and the PR rejected.

Apologize for this mess.

slandelle:
No pro, thanks for your feedback!

PR #2592, opened 2015-03-02
fix(SSE and wsAwait check): SSE blocking check does not stop after timeout

ctranxuan:
The wsAwait check is not triggered and thus won't never timeout as long as the SSE server don't send at least one event. This comes from the fact that the SseHandler must send an "OnOpen" to the SseActor when the SseHandler connection received an "onConnectionOpen" and not when the SseHandler received a status from the SSE server ("onStatusReceived").


ctranxuan:
Close the PR (see comment #2591)

PR #2610, opened 2015-03-13
Updated feeder tests

LeonardMeyer:
Package now covered at 100% 

Wait to merge though, false positive slipped through. Will fix.


slandelle:
Thanks

PR #2621, opened 2015-03-17
Fix gatling configuration not being resolved.

knshiro:
This bug prevented from using environment variables in `gatling.conf` or any other dynamically resolved variables.


slandelle:
Thanks!

PR #2629, opened 2015-03-21
Update randomString example

ZeissS:
`Random.randomString(length)` returns non-ascii characters which will probably not work for most use-cases. 

I replaced it with `Random.alphanumeric.take(20).mkString(start,end,separator)` to only contain alphanumeric chars.

### DIFF
--- a/src/sphinx/session/code/Feeders.scala
+++ b/src/sphinx/session/code/Feeders.scala
@@ -7,7 +7,7 @@ class Feeders {
   {
     //#random-mail-generator
     import scala.util.Random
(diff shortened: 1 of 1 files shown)


slandelle:
Good idea, but your code is wrong ;)
I'll merge then fix.

slandelle:
PS: your mistake is that mkString signature is `(start,separator,end)` not `(start,end,separator)`.

ZeissS:
Thanks!

Sorry for the mistake, in my local script I used only the prefix and just modified it slightly for your docs.

slandelle:
No pro, thanks for contributing.

Issue #2626, opened 2015-03-19
JMS: add support for receive timeout


soudmaijer:
Would it be possible to add support for setting a receive timeout for JMS reqreply actions.

Possibly somewhere here:

https://github.com/gatling/gatling/blob/master/gatling-jms/src/main/scala/io/gatling/jms/JmsReqReplyAction.scala#L68

Maybe I will contribute when I have some spare time.

slandelle:
Yes, contribution would be very welcome! :)

slandelle:
Backported on 2.1.

soudmaijer:
Awesome! Thanks!

PR #2630, opened 2015-03-21
JMS: add support for receive timeout

pwielgolaski:
implementation of #2626


slandelle:
Thanks a lot @pwielgolaski , you rock!

Issue #2638, opened 2015-03-25
Core: Better support for writing functional specs


constantijn:
Copy from https://groups.google.com/forum/#!topic/gatling/_BqDQFxXKz8

---------

On my project we're building a REST API and when it comes to testing we use ScalaTest for most of the unit tests, however at the end of the day we also want to know our API works properly and we decided to use Gatling to write some functional specs against our API.

It works quite well for this purpose but i had to deal with a couple of gotchas (by default failed tests don't fail the build, SBT plugin had a bug that i already submitted a fix for etc.) 
In the end after abstracting away all the boilerplate into the GatlingFunSpec class my tests ended up looking something like this:

```scala
class StuffHandlerTest extends GatlingHttpFunSpec {

  override val baseURL = "http://localhost:7654/api"
  override val httpConf = super.httpConf.header("MyHeader", "MyValue")

  spec {
    http("GetAll")
      .get("/stuff")
      .check(stuff exists)
  }

  spec {
    http("GetStuffFromCountryToCountry")
      .get("/stuff/NL/NL")
      .check(stuff.count is 1)
      .check(fromCountry is "NL")
      .check(toCountry is "NL")
  }

}

object StuffHandlerTest {

  def stuff = jsonPath("$.stuff")
  def fromCountry = jsonPath("$.stuff[0].fromCountryCode")
  def toCountry = jsonPath("$.stuff[0].toCountryCode")

}
```

Would you guys be interested in a pull request for my GatlingFunSpec implementation? I realize it's nothing new per se, but it would make it quite a bit easier for people to get started writing functional tests using Gatling.

It would require some cleaning up from it's current implementation but I'd be happy to do it (write tests, a manual page,blog post, etc.) and submit a pull request if this seems interesting to you.

Constantijn Visinescu

constantijn:
I have a first implementation ready. I'd like to nail down the implementation first before I start writing docs and submitting pull requests ;)

https://github.com/constantijn/gatling/compare/master...constantijn:gatling-funspec

I have two questions to the devs:

What would be the correct module to implement this functionality in?

gatling-main would make the most sense because that's where the simulation class this is based on lives, but it doesn't have access to gatling-http. gatling-http would sorta make sense, but I'm not sure it belongs there. I have it in gatling-testframework now which makes most sense namewise, but rightnow it only contains classes realted to the sbt testframework implentation

How would you like me to write a test for this?

[...]

PR #2641, opened 2015-03-26
Core: Better support for writing functional specs #2638

constantijn:
Core: Better support for writing functional specs #2638

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/funspec/GatlingFunSpec.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/funspec/GatlingFunSpec.scala
@@ -1,0 +1,30 @@
+package io.gatling.core.funspec
+
+import io.gatling.core.Predef._
+import io.gatling.core.action.builder.ActionBuilder
+import io.gatling.core.config.Protocol
+import io.gatling.core.structure.ChainBuilder
+
+import scala.collection.mutable.ListBuffer
+
+abstract class GatlingFunSpec extends Simulation {
+
+  /** Set the protocol configuration used by gatling to execute all specs in this class */
+  def protocolConf: Protocol
+
+  /** Add a spec to be executed */
+  def spec(actionBuilder: ActionBuilder) = specs += actionBuilder
+
+  private[this] val specs = new ListBuffer[ActionBuilder]
+
+  private[this] lazy val testScenario = scenario(this.getClass.getSimpleName)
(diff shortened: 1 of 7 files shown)


constantijn:
Duplicate issue of #2638

gatling-core/src/main/scala/io/gatling/core/runner/Runner.scala
```diff
@@ -47,6 +48,8 @@ class Runner(selection: Selection)(implicit configuration: GatlingConfiguration)
       GatlingActorSystem.start()
       val simulation = simulationClass.newInstance
 
+      if (simulation.isInstanceOf[GatlingFunSpec]) simulation.asInstanceOf[GatlingFunSpec].setupRegisteredSpecs
```
slandelle: pattern matching?

gatling-core/src/main/scala/io/gatling/core/funspec/GatlingFunSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  /** Add a spec to be executed */
+  def spec(actionBuilder: ActionBuilder) = specs += actionBuilder
+
+  private[this] val specs = new ListBuffer[ActionBuilder]
+
+  private[this] lazy val testScenario = scenario(this.getClass.getSimpleName)
+    .exec(ChainBuilder(specs.reverse.toList))
+
+  private[core] def setupRegisteredSpecs = {
```
slandelle: I was fine with DelayedInit :)
It's deprecated but won't be dropped until there's a replacement, so...

constantijn: The warnings were polluting my shiny warning-less scala 2.11 project :)
Figured i'd fix it while I was at it.

gatling-core/src/main/scala/io/gatling/core/runner/Runner.scala
constantijn: I thought of pattern matching, but I figured it was 3 lines of code where one will do. 
But I'll go change it since pattern matching is more scala-y anyway :)

constantijn: Fixed :)

slandelle:
Could you squash and rebase, please?

PR #2647, opened 2015-03-27
JMS: refactoring of JMSClient creation, to make possible to switch implementation

pwielgolaski:
it is starting point for alternative implementation of JMSClient, I'm thinking to make one that has pool of connections in near future


slandelle:
Looks like a good idea.

PR #2701, opened 2015-05-01
Recorder: Privatize most Recorder classes

pdalpra:
This basically only leave public the classes that are used to run/configure the Recorder.


slandelle:
LGTM

Issue #2700, opened 2015-05-01
Core: revert back to supporting double percentile assertions


slandelle:
See #2357.

We'd probably need 2 branches of `AssertionWithPathAndTarget` one for Ints (counts) and one for Doubles (percents).

@pdalpra WDYT?

slandelle:
Superseded by #2702

PR #2702, opened 2015-05-01
Core: Adding per-million support to assertions

djasonpenney:
When measuring the success rate in a Gatling test, the existing "percent" function is insufficient.  Most of my services require a maximum failure rate of one per thousand.

This pull request extends the "percent" concept with a per-million extension.  For my purposes, a success rate of 999000 (or a failure rate of 1000) is sufficient.  I believe that one per million is probably enough for almost any Gatling use case.


slandelle:
@djasonpenney Thanks, that's a good idea, and better than #2700. I suspect you're working with @infomaven? ;-)

djasonpenney:
Thank you, yes, infomaven and I are coworkers. Best regards.

On Fri, May 1, 2015 at 10:32 AM, Stephane Landelle <notifications@github.com

PR #2741, opened 2015-06-01
HTTP: Make Http Protocol localAddress take an Expression

pdalpra:


slandelle:
I wonder if you could reduce allocations here.

pdalpra:
We could easily make them function values instead of methods.

slandelle:
Higher order functions wouldn't reduce allocation :)

pdalpra:
Do you prefer that i switch back to the previous form and drop `ExpressionRequestConfigurator` ?

slandelle:
Yeah. Let's do that in 2 steps: first provide the feature, then let's see if we can improve.

pdalpra:
Updated the PR to include only the feature.

slandelle:
Thanks mate!

pdalpra:
You're welcome :)

PR #2819, opened 2015-09-23
Update link to maven plugin source code

edwag:
The link to the source code of the gatling maven plugin pointed to https://github.com/gatling/gatling-maven-plugin, which I believe is not current. I've updated the link to point at https://github.com/gatling/gatling-maven instead.


slandelle:
Thanks!

PR #2832, opened 2015-10-09
debugging doc page

aidylewis:


src/sphinx/general/debugging.rst
```diff
@@ -0,0 +1,61 @@
...
+If we wanted to print all values within a session we can just use this syntax:
+
+.. code-block:: scala
+
+    .exec { session => println(session); session }
+To print a single session value use syntax akin to this:
+  
+.. code-block:: scala
+
+    .exec({session => println(session("addComputer").as[String]); session })
```
slandelle: Actually, we have a trick there: Scala code samples are actually extracted from real classes that get compiled.
This way, we can make sure our samples are right, and up-to-date when we change something.
See `includecode` directive in [quickstart.rst](https://raw.githubusercontent.com/gatling/gatling/master/src/sphinx/quickstart.rst)  and how it works with [QuickStart.scala](https://github.com/gatling/gatling/blob/master/src/sphinx/code/QuickStart.scala).

Also, this code would be more simple the following way:

```scala
exec { session =>
  println(session("addComputer").as[String])
  session
}
```

I'd rather not encourage beginners to use `;` in Scala, it would bring more confusion.

src/sphinx/general/debugging.rst
```diff
@@ -0,0 +1,61 @@
...
+  
+.. code-block:: scala
+
+    .exec({session => println(session("addComputer").as[String]); session }) 
+
+Logback
+=======
+
+There should be a default logback.xml file in the Gatling conf directory. 
+Add a file appender, similar to this:
```
slandelle: Adding a file appender is not necessary for debugging.

aidylewis:
OK. Do you want to reject the pull request and I'll make the necessary changes? Is the general directory OK for the file(s) or should I create another one? Once I get the hang of this, I will look to see if I can improve other parts of the documention.

src/sphinx/general/debugging.rst
aidylewis: I added a file appender, so users could remove the none HTML response and view the page in a browser.

aidylewis:
I've squashed the two commits and given a new PR.

slandelle:
@aidylewis Finally taking the time to review, sorry for being this late. As you can see, your PR doesn't compile. Could you please fix?

src/sphinx/general/code/ComputerWorld.scala
```diff
@@ -0,0 +1,72 @@
...
+    .exec(http("getComputers")
+    .get("/computers")
+      .check(
+        status.is(200),
+        regex("""\d+ computers found"""),
+        css("#add", "href").saveAs("addComputer")))
+
+    //#print-all-session-values
+    .exec { session =>
+      println(session(session)
```
slandelle: `println(session)`

Behold the power of compiled documentation samples.

aidylewis:
OK. I will look into it,

src/sphinx/general/code/ComputerWorld.scala
slandelle: @aidylewis Could you please fix?

slandelle:
@aidylewis Could you please fix the small error so it can be shipped, please?
Sorry for the delay :(

aidylewis:
Forgot all about this. I will create a new fork.

slandelle:
I can deal with it myself, but you wouldn't appear as the committer. Let me know if it matters.

aidylewis:
Doesn't matter. Not out for the glory.

slandelle:
OK :)
Will do then.

PR #2834, opened 2015-10-15
Allow to change the escapeChar of a csv feeder

henri-tremblay:
This should do the trick. However, I must say that being fully compliant with the RFC as the default would have make more sense but would break backward compatibility.

Also, it might be possible that specifying no escape char instead of using \0 is cleaner.

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/feeder/FeederSupport.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/feeder/FeederSupport.scala
@@ -30,18 +30,19 @@ trait FeederSupport {
   implicit def array2FeederBuilder[T](data: Array[Map[String, T]]): RecordSeqFeederBuilder[T] = RecordSeqFeederBuilder(data)
   implicit def feeder2FeederBuilder[T](feeder: Feeder[T]): FeederBuilder[T] = FeederWrapper(feeder)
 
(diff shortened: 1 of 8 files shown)


slandelle:
Your code doesn't compile... :)

Then, you didn't get me:
* I'm fine with sticking to RFC as the default behavior. We'll just have to mention it in 2.2.
* Please:
  * change `separatedValues` to accept a new parameter of type Option[Char]
  * overload other methods such as `csv` with to it can also take a new parameter of type Char. You'll probably have to overload several times because of the `rawSplit` default value (can't have more than one).
  * compile and pass the tests

henri-tremblay:
I made it compliant with the RFC by default. I've also started to implement it with an Option but found myself blocked by my lack of Scala knowledge. So I'm sorry but I'll have to let you fill the holes.

slandelle:
@henri-tremblay Back from travel.

Now that I can see the whole code, I realize that passing an Option would make duplicating all DSL methods many times because of the parameter default values. And we can't expect standard users to pass an Option themselves.
It would be cleaner to have a builder pattern for passing all those options, but sadly, it would break some other use cases, for example people using a parser to gain direct to parsed data (records field). So, it would make them change their code and explicitly call the final build method. That's maybe something we'll do in a future major version if proven useful, but we'll now stick with the current design.

I'm going to pick from your PR, and direct pass a `Char`, with a `0` default value (which is the noop default value Jackson uses underneath).

Thanks

slandelle:
Closed by #2833

Issue #2816, opened 2015-09-19
SBT Plugin: avoid collision with unit tests


pdalpra:

slandelle:
Reopening, see https://github.com/gatling/gatling/issues/2763#issuecomment-148681035

pdalpra:
@slandelle can it be closed ?

slandelle:
Absolutely. Tahnks!

PR #2847, opened 2015-11-08
Fix SBT Plugin documentation

pdalpra:


pdalpra:
@slandelle should this close #2816 (again) ?

slandelle:
Is there still a point in having gatling-it?

pdalpra:
I think there is: you may want to split short lived simulations ("unit" tests) and long-lived ones (load tests) so having the ability to have separate folders seems useful IMHO

slandelle:
LGTM, thanks!

PR #2874, opened 2015-12-14
Add statsD protocol support for realtime metrics

fromanator:
Adding support for pushing metrics to statsD instead of directly to graphite. Reasons for doing so are:

* The Gatling App no longer has to handle calculating the metrics in a buffer to send to graphite, because statsD will do these calculations for us.
* Allows for a more flexible and scalable infrastructure for handling realtime metrics over going to a graphite server directly.
* We can down sample some metrics that are normally at a high volume to control bottlenecks in reporting metrics. StatsD takes the sample rate into account when it derives metrics based on the events like P99 response times.

### DIFF
--- a/gatling-core/src/main/resources/gatling-defaults.conf
+++ b/gatling-core/src/main/resources/gatling-defaults.conf
@@ -115,5 +115,11 @@ gatling {
       bufferSize = 8192          # GraphiteDataWriter's internal data buffer size, in bytes
       writeInterval = 1          # GraphiteDataWriter's write interval, in seconds
     }
+    statsd {
+      light = false       # only send the all* stats
+      host = "localhost"  # The host where the statsd server is located
+      port = 8125         # The port to which the statsd server listens to (8125 is the default)
+      protocol = "udp"    # The protocol used to send data to statsd (currently supported : "tcp", "udp")
+    }
   }
 }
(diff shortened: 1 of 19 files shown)


slandelle:
@fromanator I'd like to get in touch. Could you please send me an email at slandelle@gatling.io?

Limess:
Is there any reason why this wasn't accepted?

fromanator:
@Limess so unfortunately the conversation was moved to email instead of this PR, but for the sake of openness (this is an open source project) I'll post the details:

I work for [Target Corporation](https://en.wikipedia.org/wiki/Target_Corporation) linking to Wikipedia in case you are not familiar since we only sell in the United States (briefly in Canada). Anyway I did this work to integrate into our infrastructure where we use statsD to aggregate our metrics for extra flexibility in which backends (like Graphite) we want it to go to. After implementing it I also wanted to contribute back so others could use it too.

However the core development team from now Gatling Corp want to do the following:
"
- Everything related to running the tests and performing load is open-source. So is whatever that's be published and distributed for free so far: the reports, the existing Jenkins plugin, the Graphite integration... This scope will keep on being developed actively. For example, upcoming 2.2 will bring OpenSSL and native Linux epoll support for even more performance.
- Every new feature related to integrating with other tools or providing more analytics will come into our commercial product.

"

So basically if you want integration more than just the generated report and Graphite, you would need http://gatling.io/#/services/frontline . 

I can understand the viewpoint of not wanting to support every integration out there as a core part of the project, since that means more work to keep them up to date. However I have been thinking that a good compromise would be to provide a system where you can plug in a 3rd party [DataWriter](https://github.com/gatling/gatling/blob/master/gatling-core/src/main/scala/io/gatling/core/stats/writer/DataWriter.scala) to use.

Limess:
Thanks for the context @fromanator, that clears things up.

PR #2880, opened 2015-12-30
Gather messages from all Failures

Mishail:
Current implementation of `Check.check` ignores all Failures except the 1st one. We could use `mapError` to combine all the messages from failures into the resulting `Failure`


slandelle:
Hi @Mishail,

This has been asked for several times, and hasn't been implemented because it's actually more complex (or we would have done it already :) ).

* You can't use `\n` as a separator, as those errors are logged into a TSV file. There would be other solutions (like reencode in Base64), but what about logging performance 
* We don't want to go with something like Scalaz's Validation that would let use pile up errors, because it would both hurt performance and be too complex for most Gatling users
* How would we display those errors in every Gatling output: reports, console, Graphite?
* They would have to split back. Perf impact?

In short, such change would have to because carefully designed and perf impact properly measured.

Suggestions welcome.

PR #2904, opened 2016-02-08
Provide a factory for Simulation classes instantiation.

aoboturov:
Some projects which use Gatling could have different need when it comes to instantiation of simulation objects. At the moment it is not parametric. I propose to change it.

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/Gatling.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/Gatling.scala
@@ -19,11 +19,13 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util.{ Failure, Try }
 
+import io.gatling.app.Gatling.SimulationFactory
 import io.gatling.app.cli.{ StatusCode, ArgsParser }
 import io.gatling.commons.util.{ Ga, StringHelper }
 import io.gatling.commons.util.TimeHelper._
 import io.gatling.core.config.GatlingConfiguration
 import io.gatling.core.controller.ControllerCommand
+import io.gatling.core.scenario.Simulation
 import io.gatling.core.stats.writer.RunMessage
 
 import akka.actor.ActorSystem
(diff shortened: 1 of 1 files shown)


slandelle:
In order to get a change, you're expected to explain your use case/motivation so we're convinced this change makes sense and will benefit the project and other users.

No offense, but introducing a kind of hook for only one user so he can hack the rest of a feature on his own without ever re-contributing doesn't benefit the project at all.

aoboturov:
@slandelle you're right.

The use case: we have a number of backends implementing the same API and we would like to have some parametrization which would not be convenient to express in scala/java gatling dsl.

Gatling is embedded as a part of our framework and there is routing based on parametrization. Initially we had a base class for the simulation and then a specialization for each backend, i.e. N*M classes, where N is the number of simulations and M is the number of backends. To have less classes and simplify our config we have introduced this factory to have one simulation class with a constructor containing parameters, which obviously would not work in current approach. The other way to do it - is to write macro which would expand to generated classes.

Having this use case in mind, one could have a standalone app which is deployed in cloud and without further re-deployment we could have multiple simulations running against our backend without recompile our simulator. With such a factory, parametrization could be passed in runtime and we would not have to pay for re-deployment which in AWS case could take 15 mins for some configurations.

slandelle:
I really fail to see why you can't achieve that with `Gatling.fromArgs`. Plus, the Gatling class is not a public API, only methods on the companion.

aoboturov:
`Gatling#fromArgs` would permit to have variable parameters via  `case class GatlingConfiguration( ... , config: Config)` which is handy for basic-parametrization of simulations, like the number of users, requests, delays etc.

Our case is more about having something like:

```
trait BackendAdaptedQuery { this: Simulation =>
  def createAdapter: BackendAdapter
  val adapter: BackendAdapter = createAdapter
}

class SomeSimulation(backendAdapter: BackendAdapter, ...) extends Simulation with BackendAdaptedQuery {
...
// use adapter for query adaptation to specific backend.
...
}
```

Here `adapter` would do some backend-specific query adaptation. We can not have it specified as a part of TypeSafe config - it's not JSON-serializable object for instance, and it is not supposed to be such an object.

gatling-app/src/main/scala/io/gatling/app/Gatling.scala
```diff
@@ -48,11 +52,11 @@ object Gatling {
 
     implicit val configuration = GatlingConfiguration.load(overrides)
 
-    new Gatling(selectedSimulationClass).start.code
+    new Gatling((v) => v.newInstance())(selectedSimulationClass).start.code
   }
 }
 
-private[app] class Gatling(selectedSimulationClass: SelectedSimulationClass)(implicit configuration: GatlingConfiguration) {
+private[app] class Gatling(simulationFactory: SimulationFactory = _.newInstance())(selectedSimulationClass: SelectedSimulationClass)(implicit configuration: GatlingConfiguration) {
```
slandelle: No reason to use currying here

slandelle: This class is package private. Users are only supposed to use factories in the companion. Provide a public API.

aoboturov:
Done. Btw, do you know if it's my changes that are failing the build? In Travis I can see timeouts, but normally those changes are not supposed to change anything because default values would still allow code to compile and have the same run-time behavior.

slandelle:
cherry-picked. It will be streamed down to the public repo as soon as we can.

> Btw, do you know if it's my changes that are failing the build?

akka-testkit timing failures when Travis' hosts are swamped...

aoboturov:
Thanks a lot for quick iteration :)

PR #2912, opened 2016-02-12
We need timeouts to be recorded as failures

kiggundu:
For now... this modifications ensures that if a timeout is hit the request is logged as a failure along with all currently in flight requests. 
Need to improve solution to track the timed-out request and fail only that one however at the moment simply setting the timeout to a realistic value will ensure that only the log lasting call will still be in the sent messages map


slandelle:
Cherry-picked, thanks

Issue #2913, opened 2016-02-12
JMS: Responses not properly handled when multiple actions use the same replay queue


kiggundu:
It looks like the JMS plugin when used with activeMQ behaves unpredictably, randomnly dropping requests and responses. I've attached a sample test script that will reproduce the issue. What happens is that though the messages are successfully submitted and the responses are successfully dequed the JMS client fails to consistently pass this on to the gatling reporting framework and the tracker simply ticks on waiting to be notified
If all the tasks are commented out of the scenario appart from one then all works fine because only one thread is created. I believe this has something to do with JMS session thread affinity but not certain.

Simulation attached below:
[JMSConcurrencyIssue.txt](https://github.com/gatling/gatling/files/128347/JMSConcurrencyIssue.txt)

slandelle:
```scala
  private def initExecution(title: String): JmsRequestBuilder = {
    jms(title)
      .reqreply
      .queue("abraham.simulation.in")
      .replyQueue("abraham.simulation.out")
      .textMessage(stringToExpression("Client asking server to: " + title))
      .property("transactionId", java.util.UUID.randomUUID.toString.toUpperCase())
  }
```

Is it intended that transactionId is static for a given exec type? Are you sure you don't want to pass an Expression instead (one transactionId per message sent)?

slandelle:
Your problem is that all your actions read and write from the same queues.

Trackers keep track of sent ids and are in charge of performing correlation. The thing is there's one tracker per action. But as you have several of them, and they all read from the same queue, there's a very good chance that a given response gets picked up by a different tracker than the one that kept track of the request, hence correlation never happens.

slandelle:
IMO, this is a design issue. There should be only one tracker per replyDestination, or a single one that would use the replyDestination in the tracking key.

kiggundu:
Hi Stephane,

Thanks for looking into this. Sounds like a change needs to be made
within the JMS client to ensure that there is one tracker per queue. If
that's so are you able to advise on when a patch might be available?
We'll stand down if that's pretty soon, if not we can take a stab at
fixing it.

rgds
Abraham

Stephane Landelle:

slandelle:
Fixed

kiggundu:
Brilliant,
Many thanks for the quick turnaround on this one Stephan.

kiggundu:
We're keen to integrate with your fix as soon as possible. When will you be in a position to push the changes?

slandelle:
[...]

PR #2918, opened 2016-02-18
Gatling JMS client concurrency issue

kiggundu:
The fix provided for issue #2913 in build 2.2.0-E20160217 worked fine agains the testcase provided in that ticket however when the number of concurrent users was increased the same unpredictable concurrency behaviour returned.
Looking at the fix it looks like the tracker was placed into the executing context which meant that a new one was created per context and thus per user. This pull request is a solution we coded which seems to be working fine for us under heavy user loads.

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/request/builder/RequestExpressionBuilder.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/request/builder/RequestExpressionBuilder.scala
@@ -79,7 +79,7 @@ abstract class RequestExpressionBuilder(commonAttributes: CommonAttributes, http
         //
         // [fl]
         _.setNameResolver(nameResolver)
-      }
+    }
 
   // FIXME resolve proxy presence once
   private def configureProxy(requestBuilder: AhcRequestBuilder, uri: Uri): Validation[AhcRequestBuilder] = {
--- a/gatling-jms/src/main/scala/io/gatling/jms/action/JmsReqReplyAction.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/action/JmsReqReplyAction.scala
@@ -67,6 +67,7 @@ class JmsReqReplyAction(attributes: JmsAttributes, protocol: JmsProtocol, tracke
               tracker ! MessageReceived(messageMatcher.responseID(msg), nowMillis, msg)
               logMessage(s"Message received ${msg.getJMSMessageID}", msg)
             case _ =>
+              tracker ! FailureReceivingMessage()
               throw BlockingReceiveReturnedNull
           }
         }
--- a/gatling-jms/src/main/scala/io/gatling/jms/action/JmsRequestTrackerActor.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/action/JmsRequestTrackerActor.scala
@@ -48,6 +48,12 @@ case class MessageSent(
  */
 case class MessageReceived(responseId: String, received: Long, message: Message)
 
+/**
+ * Advise actor that something went wrong while receiving a message
+ * This includes timeouts
+ */
+case class FailureReceivingMessage()
+
 object JmsRequestTrackerActor {
   def props(statsEngine: StatsEngine) = Props(new JmsRequestTrackerActor(statsEngine))
 }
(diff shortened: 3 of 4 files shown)


slandelle:
> Looking at the fix it looks like the tracker was placed into the executing context which meant that a new one was created per context and thus per user.

Absolutely not. There's now only one single instance of `JmsRequestTrackerActor` per scenario. Message tracking then account for receiving queue.

No offense, but your proposed fix is wrong. It completely missed the whole point of Akka actors and what they guarantee in regard to the [JMM](http://doc.akka.io/docs/akka/2.4.2/general/jmm.html). There's absolutely no need for this `RequestResponseCorrelator` and all those Atomic and Concurrent structures, and all the more this scheduled `Runnable`.

For now, the only thing I see that could go wrong with thread-safety/memory visibility would be if your JMS implementation (which is, BTW?) was mutating the tracked `javax.jms.Message` after the `MessageProducer` sends them, or after the `MessageConsumer` receives them.

Please elaborate on your issue and the JMS implementation you're using, and even better, provide a reproducer.

kiggundu:
Definitely no offense taken, I believe we all have vested interest in this being the best working solution possible. I must admit you're right, I have minimal experience with Akka much less its guarantees.

I have re-run the modified test-case that failed against your fix and completely failed to reproduce the failure (... no room for any more F's in that sentence).  I may well have not 'controlled' the test properly meaning the failure could have been caused by my queues not responding. Will ignore our changes at your advise for and keep you posted if that changes.

slandelle:
> I have minimal experience with Akka much less its guarantees

TLDR;
Actor state is threadsafe as long as mutations only happen by message passing.

>  There's now only one single instance of JmsRequestTrackerActor per scenario

Correcting myself, `JmsRequestTrackerActor` is a proper global singleton. So using the same queue from multiple scenarios is safe.

> keep you posted if that changes

Please do. If you can get a reproducer, I'll gladly take a stab at it!

kiggundu:
We continued to observe issues where though all messages were sent and replies picked up from the queues the internals of the JMS plugin sometimes tend to loose the replies. This is sporadic... i.e. sometimes it wrks and sometimes it doesn't so there is no real testcase i can submit. On looking again at the code I am still convinced that though there is only one tracker as per your fix and the advice above it is being hit concurrently by multiple threads yet it is not thread-safe wrt the sentMessages and receivedMessages fields.
 I produced the attached logs for the code at [this diff](https://github.com/gatling/gatling/compare/master...kiggundu:concurrencyFix?expand=1
)

[loadtest.usingtracker.log.txt](https://github.com/gatling/gatling/files/139770/loadtest.usingtracker.log.txt):
 This is with the code tracking the requests and responses within the JmsRequestTrackerActor as per the current implementation

[loadtest.externalsingleton.log.txt](https://github.com/gatling/gatling/files/139769/loadtest.externalsingleton.log.txt):
 This is with the code as per the stated diff (i.e. with the tracking variables synchronised. Only put them in there own singleton for clarity so don't think that's necessary)

because the sentmessages and receivedMessages fields are not protected you can see that the first logs are full of race conditions while the second are not. I would expect this to occur even if there was only one instance of the tracker unless i have greatly mis-understood the guarantees clarified above.

Please advise

slandelle:
> because the sentmessages and receivedMessages fields are not protected you can see that the first logs are full of race conditions while the second are not.

Sorry, your test is broken if you're accessing the `sentmessages` and `receivedMessages` Actor  fields directly. Please check again the [Akka JMM guarantees](http://doc.akka.io/docs/akka/2.4.2/general/jmm.html): you must NEVER do that: mutations and memory visibility are only safe through message passing.

If you think you have an issue, please provide a demo app. But then, also consider the bug could be in your test or system under test. Are you sure that your correlationIds are random enough?

slandelle:
PS: I'm turning `sentMessages` and `receivedMessages` visibility to private :)

kiggundu:
That's a good move. 
We have definitely seen the behaviour again using the provided fix but on a more sporadic basis. Got 404's trying to access the fix's repository so had no option but to investigate using master but it could well be that our test has thus conflated both the issue that was fixed and another issue observed in the fixed code.

The "The actor subsequent processing rule" seems to be rule out my concurrency suspicions however, the fact is that we are still getting runs that appear to have internally lost replies consequently labeling them as errored/timed-out when it is obvious by looking at the queues that they were all sent back well before the timeout.

Can only really investigate further at our end if we have access to the fix repository though.

slandelle:
Please get in touch at contact at gatling dot io

PR #2926, opened 2016-02-25
Add a parameter to choose group timings displayed in report. Close #2924.

### LINKED ISSUE #2924: Show group duration in main report table instead of cummulated response time
For some of my simulations, group duration is more important than cumulated response time. I propose to add a switch in configuration to choose between two timings.

For the records, it is close to #2091 .

mxbossard:
By default, keep previous behaviour: group timings displayed are
cumulated response times. This can be replaced in config to display
group duration instead.
Close #2924.

### DIFF
--- a/gatling-core/src/main/resources/gatling-defaults.conf
+++ b/gatling-core/src/main/resources/gatling-defaults.conf
@@ -43,6 +43,7 @@ gatling {
   charting {
     noReports = false       # When set to true, don't generate HTML reports
     maxPlotPerSeries = 1000 # Number of points per graph in Gatling reports
+    useGroupDurationInsteadOfCumulatedResponseTime = false  # Switch group timings from cumulated response time to group duration.
     indicators {
       lowerBound = 800      # Lower bound for the requests' response time to track in the reports and the console summary
       higherBound = 1200    # Higher bound for the requests' response time to track in the reports and the console summary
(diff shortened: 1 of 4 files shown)


gatling-core/src/main/resources/gatling-defaults.conf
```diff
@@ -43,6 +43,7 @@ gatling {
   charting {
     noReports = false       # When set to true, don't generate HTML reports
     maxPlotPerSeries = 1000 # Number of points per graph in Gatling reports
+    useGroupDurationInsteadOfCumulatedResponseTime = false  # Switch group timings from cumulated response time to group duration.
```
slandelle: mmmm, trying to come up with a shorter name...

mxbossard: It's a switch between two kind of timings for groups. 
Firstly, I wanted to let the possibility to display all timings in report like this : 
mode 1 : cumulatedResponseTime
mode 2 : groupDuration
mode 3 : cumulatedResponseTime | groupDuration

It's more complex to implement, but we could begin by implementing mode 1 and 2. And use a switch paramater name like : groupTimingsDisplayMode = 1

slandelle: Being able to handle both duration and cumulated response time together would require a lot of work and consideration:
* how to deal with the stats table on the Global page?
* what about assertions on groups, the API only let you target one metric?

I'm completely fine with the implementation you suggest. My only issue is with the option name, which meaningful, but very long. What about `useGroupDurationMetric`? The explaining comment looks sufficient to me so users know what they're doing.

Then, there's one thing missing from your PR: documentation. Could you please add some documentation, probably in `reports.rst`?

Thanks!

mxbossard:
I close this one cause I opened a new pull request : #2929 (instead of adding a second commit to the original branch).

slandelle:
@mxbossard Why don't you just rebase/amend on your local clone and then push force on your fork?

PR #2931, opened 2016-03-01
Feature: conditional check. Close #2927

### LINKED ISSUE #2927: Core : Conditionnal check
For my use case testing a Vaadin app, I need a mean to perform a check only if one condition against the response or the session is True.

I ended up by developping a check wrapper which take a function : (session, response) => boolean and a nested check in parameter and perform the nested check if the function return true.
I am not sure it has no impact on performance, but I need it cause my scenarios are highly coupled with the server responses.

Here is a simple use case for this feature : I check all responses with a jsonpath check, but some responses payload are not formated in json. So I supplied a jsonSecured check wich is a ConditionnalCheck with a condition on the reponse content-type header, to perform the jsonpath nested check only if the response contains a JSON payload.

Are you interested in this feature ?

mxbossard:
Here is a fist draft of the implementation.
I think it's nice. It compile, but it's not tested yet.

Close #2927 .

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
@@ -16,6 +16,8 @@
 
 
 
+
+
 package io.gatling.core.check
 
 import java.util.UUID
@@ -28,29 +30,27 @@ import io.gatling.core.session.Expression
 
 object ConditionalCheck extends LazyLogging {
   
-  implicit def conditionalCheckBuilder2Check[C <: Check[R], R](conditionalCheckBuilder: ConditionalCheckBuilder[R, C]) = conditionalCheckBuilder.build
-  
-  trait CheckWrapper[R, C <: Check[R]] {
-    def wrap(check: Check[R]): C
+  implicit def conditionalCheckBuilder2Check[R, C <: Check[R]](conditionalCheckBuilder: ConditionalCheckBuilder[R, C]): ConditionalCheck[R, C] = conditionalCheckBuilder.build
+
+  trait ConditionalCheckWrapper[R, C <: Check[R]] {
+    def wrap(check: ConditionalCheck[R, C]): C
   }
 
-  def checkIf[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = cw.wrap(new ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, None))
+  def checkIf[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: C)(implicit cw: ConditionalCheckWrapper[R, C]): C = cw.wrap(ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, None))
   
-  def checkIfOrElse[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: Check[R])(elseCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = cw.wrap(new ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, Some(elseCheck)))
+  def checkIfOrElse[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: C)(elseCheck: C)(implicit cw: ConditionalCheckWrapper[R, C]): C = cw.wrap(ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, Some(elseCheck)))
   
-  def checkIf[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean])(thenCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = 
-    cw.wrap(new ConditionalCheckBuilder((r: R,s: Session) => condition(r, s), thenCheck, None))
+  def checkIf[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean])(thenCheck: C)(implicit cw: ConditionalCheckWrapper[R, C]): C = 
+    cw.wrap(ConditionalCheckBuilder((r: R,s: Session) => condition(r, s), thenCheck, None))
         
-  def checkIfOrElse[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean])(thenCheck: Check[R])(elseCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = 
-    cw.wrap(new ConditionalCheckBuilder((r: R,s: Session) => condition(r, s), thenCheck, Some(elseCheck)))
+  def checkIfOrElse[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean])(thenCheck: C)(elseCheck: C)(implicit cw: ConditionalCheckWrapper[R, C]): C = 
+    cw.wrap(ConditionalCheckBuilder((r: R,s: Session) => condition(r, s), thenCheck, Some(elseCheck)))
 
   case class ConditionalCheckBuilder[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean], thenCheck: C, elseCheck: Option[C]) {
-    def build: Check[R] = {
-      new ConditionalCheck[R, C](condition, thenCheck, elseCheck)
-    }
+    def build = ConditionalCheck[R, C](condition, thenCheck, elseCheck)
   }
 
-  class ConditionalCheck[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean], thenCheck: C, elseCheck: Option[C]) extends Check[R] {
+  case class ConditionalCheck[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean], thenCheck: C, elseCheck: Option[C]) extends Check[R] {
 
     val checkUuid: String = UUID.randomUUID.toString
     
--- a/gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
@@ -1,0 +1,82 @@
+package io.gatling.core.check
+
+import com.typesafe.scalalogging.LazyLogging
+
+import io.gatling.core.session._
+import io.gatling.commons.validation.Validation
+import io.gatling.commons.validation.Success
+import io.gatling.core.check.extractor.Extractor
+import java.util.UUID
+import io.gatling.core.check.CheckResult
+import io.gatling.core.check.Check
+import io.gatling.core.check.DefaultFindCheckBuilder
+import io.gatling.core.check.{ Check, Extender, Preparer }
+import io.gatling.core.Predef._
+
+object ConditionalCheck extends LazyLogging {
+  
+  trait CheckWrapper[R, C <: Check[R]] {
+    def wrap(check: Check[R]): C
+  }
+
+  def checkIf[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = cw.wrap(new ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, None))
+  
+  def checkIfOrElse[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: Check[R])(elseCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = cw.wrap(new ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, Some(elseCheck)))
+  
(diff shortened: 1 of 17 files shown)


mxbossard:
The expected behaviour is : 

checkIf(true)(failureA) => failureA
checkIf(true)(successB) => successB
checkIf(false)(failureC) => passthrough (successZ)
checkIf(false)(successD) => passthrough (successZ)

I'm writting a unit test, but for now, either or both unit test or/and implementation are failing.

mxbossard:
Hi @slandelle,
I just provided my currently working impl with some Unit Test. Now we should debate over the temporary spec I supplied, and over the implementation.

I am not sure about my choices to use : 
- DefaultFindCheckBuilder
- dummyPreparer
- dummyExtractor

On the perf side, there is certainly some improvement to do.

slandelle:
Sorry for not being able to check this as fast as I'd like to. Swamped... We'll definitely ship this for 2.2. Thanks for your support and patience

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  def dummyPreparer[R]: Preparer[R, R] = (r: R) => Success(r)
+  def dummyExtractor[R]: Expression[Extractor[R, R]] = new ExpressionSuccessWrapper(new DummyExtractor[R]()).expressionSuccess
+
+  class ConditionalCheckBuilder[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean], thenCheck: C, elseCheck: Option[C]) 
+      extends DefaultFindCheckBuilder[Check[R], R, R, R](
+          (wrapped: Check[R]) => new ConditionalCheck[R, C](wrapped, condition, thenCheck, elseCheck), dummyPreparer, dummyExtractor) {
+
+  }
```
slandelle: I don't get why you need `DummyExtractor/dummyPreparer/dummyExtractor/DefaultFindCheckBuilder`. You don't want to be able to call `find` on `checkIf `.

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
```diff
@@ -0,0 +1,82 @@
...
+import io.gatling.core.session._
+import io.gatling.commons.validation.Validation
+import io.gatling.commons.validation.Success
+import io.gatling.core.check.extractor.Extractor
+import java.util.UUID
+import io.gatling.core.check.CheckResult
+import io.gatling.core.check.Check
+import io.gatling.core.check.DefaultFindCheckBuilder
+import io.gatling.core.check.{ Check, Extender, Preparer }
+import io.gatling.core.Predef._
```
slandelle: organize imports

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  def checkIf[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = cw.wrap(new ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, None))
+  
+  def checkIfOrElse[R, C <: Check[R]](condition: Expression[Boolean])(thenCheck: Check[R])(elseCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = cw.wrap(new ConditionalCheckBuilder((r: R, s: Session) => condition(s), thenCheck, Some(elseCheck)))
+  
+  def checkIf[R, C <: Check[R]](condition: (R, Session) => Validation[Boolean])(thenCheck: Check[R])(implicit cw: CheckWrapper[R, C]): C = 
+    cw.wrap(new ConditionalCheckBuilder(
+        (r: R,s: Session) => condition(r, s)
+        , thenCheck
```
slandelle: not or style: comma at the end of line. Doesn't your code gets automatically properly formatted when you build?

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
```diff
@@ -0,0 +1,82 @@
...
+    extends Check[R] {
+
+    val checkUuid: String = UUID.randomUUID.toString
+    
+    def performNestedCheck(nestedCheck: Check[R], response: R, session: Session)(implicit cache: scala.collection.mutable.Map[Any, Any]): Validation[CheckResult] = {
+      nestedCheck.check(response, session)
+    }
+    
+    def check(response: R, session: Session)(implicit cache: scala.collection.mutable.Map[Any, Any]): Validation[CheckResult] = {
+      val validationResult = condition.apply(response, session).flatMap { c => {
```
slandelle: no need for `apply`

slandelle: no need for those extra `{}`

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
```diff
@@ -0,0 +1,82 @@
...
+        if (c) {
+          logger.trace("Check: [{}] condition evaluate to true: perform nested then check.", checkUuid)
+          performNestedCheck(thenCheck, response, session)
+        } else {
+          if (elseCheck.isDefined) {
```
slandelle: OMG!!! `match`

gatling-http/src/main/scala/io/gatling/http/Predef.scala
```diff
@@ -15,8 +15,32 @@
...
 
+import io.gatling.http.check.HttpCheck
+import io.gatling.http.response.Response
+import io.gatling.http.check.HttpCheckScope
+
+import io.gatling.commons.validation.Success
+
+import io.gatling.core.session._
+import io.gatling.core.check.Check
+import io.gatling.core.check.ConditionalCheck._
```
slandelle: organize imports

gatling-http/src/main/scala/io/gatling/http/Predef.scala
```diff
@@ -15,8 +15,32 @@
...
+
 object Predef extends HttpDsl {
 
   type Request = org.asynchttpclient.Request
   type Response = io.gatling.http.response.Response
+
+  implicit object HttpCheckWrapper extends CheckWrapper[Response, HttpCheck] {
+    def wrap(check: Check[Response]): HttpCheck = new HttpCheck(check, HttpCheckScope.Body, None)
```
slandelle: That's an issue there. You can't assume scope is Body and bodyUsage is None: it depends on the check.
I suggest:
* we change HttpCheck so scope and bodyUsage are multivalues (Set)
* CheckWrapper can only wrap one given type of Check, so we have a `HttpConditionalCheckWrapper` that can access the inner checks and compute a sum of their scopes and bodyUsages

gatling-http/src/main/scala/io/gatling/http/Predef.scala
```diff
@@ -15,8 +15,32 @@
...
+  }
+
+  val HTTP_CONTENT_TYPE_HEADER_KEY: String = "Content-Type"
+  val JSON_CONTENT_TYPE_VALUE: String = "application/json"
```
slandelle: We use the Netty constants, see `HttpHeaders`

gatling-http/src/main/scala/io/gatling/http/Predef.scala
```diff
@@ -15,8 +15,32 @@
...
+  
+  def isJsonResponse(response: Response): Boolean = {
+    response.isReceived && response.header(HTTP_CONTENT_TYPE_HEADER_KEY).exists { x => x.contains(JSON_CONTENT_TYPE_VALUE) }
+  }
```
slandelle: This shouldn't be here. We should consider having such helpers on `Response`.

gatling-http/src/main/scala/io/gatling/http/Predef.scala
```diff
@@ -15,8 +15,32 @@
...
+
+  def securedJsonCheck(check: Check[Response]): HttpCheck = checkIf((response: Response, session: Session) => Success(isJsonResponse(response)))(check)
```
slandelle: Not sure about this. This is not a standard usage: YOU need this because your web framework is flawed and your API is not properly RESTful. We can have this in the documentation though.

gatling-http/src/main/scala/io/gatling/http/Predef.scala
slandelle: Plus, this has nothing to do in Predef.

gatling-http/src/test/scala/io/gatling/http/check/body/ConditionalCheckSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+import io.gatling.core.session.Session
+import io.gatling.http.HttpDsl
+import io.gatling.http.response.{ StringResponseBody, Response }
+
+import scala.collection.mutable
+
+import io.gatling.commons.validation.Success
+import io.gatling.core.check.ConditionalCheck._
+import io.gatling.http.Predef._
+import io.gatling.http.check.HttpCheck
```
slandelle: organize imports

gatling-http/src/main/scala/io/gatling/http/Predef.scala
mxbossard: Yup It has nothing to do here, I agree. It was an example of use to demonstrate the usage.

gatling-http/src/main/scala/io/gatling/http/Predef.scala
mxbossard: I knew this was a flaw, but I didn't have the big picture, so I cannot do best. I'll try dig in your way.

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
mxbossard: I will change this, my comprehension of the gatling implementation is better now.

gatling-core/src/main/scala/io/gatling/core/check/ConditionalCheck.scala
mxbossard: On this, my sbt on windows is anoying, it reformat all the sources, replacing /n by /r/n, ... Can I disable the default format and format only some files like go compiler do ?

gatling-http/src/main/scala/io/gatling/http/Predef.scala
slandelle: Typically, we add such samples in `HttpCompileTest`.

mxbossard:
I wonder if it is possible to do something like this : 
def IfCheckSuccess[R, C <: Check[R]](check: C)(cascadeCheck: C): C
How can I merge Validations ? Can we merge a Success and a Failure, 2 Success or 2 Failures ? Or do we need to define an absorbing state which may be passed ?

The spec of Cascading check could be : 
- always perform the first check
- if it succeed (or failed) => perform the cascadin check
- return the primary success (or fail) + the cascading validation

If could return 2 Success, 2 Failure, or 1 Success and 1 Failure.

gatling-http/src/main/scala/io/gatling/http/check/HttpCheck.scala
```diff
@@ -29,7 +30,7 @@ import io.gatling.http.response.{ Response, ResponseBodyUsageStrategy }
  * @param scope the part of the response this check targets
  * @param responseBodyUsageStrategy how this check uses the response body
  */
-case class HttpCheck(wrapped: Check[Response], scope: HttpCheckScope, responseBodyUsageStrategy: Option[ResponseBodyUsageStrategy])
+case class HttpCheck(wrapped: Check[Response], scope: Set[HttpCheckScope], responseBodyUsageStrategy: Set[ResponseBodyUsageStrategy])
```
slandelle: Sorry. I actually realize this is wrong.

For example, we only automatically add a status check when none is defined.
But consider the case where a checkIf sets a body check in one branch and a status one in the other one.
We would end up failing to set the default status check for the first branch!

slandelle: I want to conduct some experiments this week on the Checks design. Let me ping you once I'm done.

mxbossard: I'm sorry I don't understand the purpose of checks scope or usage strategy yet.
I let you guide me on this.

slandelle: * scope: "should we add a default status check?", "can we discard body chunks because there's no check using the body?"
* body usage strategy: "how does this check use the body so we can compute in the most efficient manner this input from the response body chunks?"

PR #2934, opened 2016-03-08
Example of run description config

mmartinsson:
Info about the new configuration of run description. Should go live together release of https://github.com/gatling/gatling-maven/pull/8.


slandelle:
Thanks

PR #2942, opened 2016-03-15
Doc: fixed typo in cheat sheet.

Senci:
SSE = Server-Sent Events, not "Server Side Events"!
Just one word exchanged. Decided to adopt the notation used in the docs/http/sse ("Server Sent Events") instead of the more widely used notation with dash ("Server-Sent Events") because of consistency reasons.


slandelle:
Facepalm! Thanks!

PR #2966, opened 2016-04-14
JMS: Mitigate memory leak caused by retaining received message duplicates

kiggundu:
We found that given 2GB of RAM gatling was failing with OOM's after a few hours. On investigating the heap-dumps we found that the JMS plugin was holding onto received messages. Further investigation highlighted the cause as due to the following:
Sometimes we send one JMS request that generates more than one JMS response to the client (i.e multiple notifications). Unfortunately the responses after the first response (which is matched with the sent message before the sent message is removed from the map) are cached by JmsRequestTrackerActor.receivedMessages and never deallocated. Multi-hour soak tests thus slowly leak memory until they burst the banks allocated to the VM. We found that the attached change reduced the leak to negligible levels.

### DIFF
--- a/gatling-jms/src/main/scala/io/gatling/jms/action/JmsRequestTrackerActor.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/action/JmsRequestTrackerActor.scala
@@ -74,6 +74,7 @@ class JmsRequestTrackerActor(statsEngine: StatsEngine) extends BaseActor {
 
   private val sentMessages = mutable.HashMap.empty[MessageKey, MessageSent]
   private val receivedMessages = mutable.HashMap.empty[MessageKey, MessageReceived]
+  private val historicallySentMessages = mutable.HashMap.empty[MessageKey, Null]
 
   // Actor receive loop
   def receive = {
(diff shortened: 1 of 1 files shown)


kiggundu:
The `java.util.concurrent.TimeoutException: Futures timed out after [2 seconds]` that caused the check failures above does not seem to be due to the changes in the pull-request.

gatling-jms/src/main/scala/io/gatling/jms/action/JmsRequestTrackerActor.scala
```diff
@@ -74,6 +74,7 @@ class JmsRequestTrackerActor(statsEngine: StatsEngine) extends BaseActor {
 
   private val sentMessages = mutable.HashMap.empty[MessageKey, MessageSent]
   private val receivedMessages = mutable.HashMap.empty[MessageKey, MessageReceived]
+  private val historicallySentMessages = mutable.HashMap.empty[MessageKey, Null]
```
slandelle: * This can grow infinitely. Please use a LinkedHashMap based LRU where you use insert order and override `removeEldestEntry`. Limit should be configurable from `gatling.conf`
* Please don't use null as  values. Use Boolean.TRUE instead.

slandelle:
Hi.

Thanks for reporting and suggesting a fix.

I'm going to address the comments myself, as we're in the process of releasing Gatling 2.2.0.

PR #2969, opened 2016-04-15
Fix typos in documentation

GabrielPlassard:
Hi,

I noticed some little typos in the documentation for 2.2, which this PR should fix (although I couldn't test it since I don't have sphinx(?) installed)

BTW, should'nt [the example for the polling](https://github.com/gatling/gatling/blob/master/src/sphinx/http/code/Polling.scala) assign the result of 
`polling.pollerName("myCustomName")`
to a val and reuse it after? Polling seems to be immutable and the other calls would not use the same pollerName. Am I missing something?

Anyway, good job on 2.2! :-)

### DIFF
--- a/src/sphinx/http/polling.rst
+++ b/src/sphinx/http/polling.rst
@@ -42,7 +42,7 @@ When you don't need to poll a request anymore, you can stop the poller:
 
 For example:
 
-.. includecode:: code/Polling.scala#pollerStart
+.. includecode:: code/Polling.scala#pollerStop
 
 .. note::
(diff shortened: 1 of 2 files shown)


slandelle:
Thanks a lot, Gabriel!

slandelle:
> BTW, shouldn't the example for the polling assign the result of polling.pollerName("myCustomName") to a val and reuse it after?

No. Giving a polling a name is only useful when you when you want several of them. They end up being stored in the session in a Map. If you don't specify one, we use an UUID and can only retrieve if there's only one entry.

GabrielPlassard:
Makes sense, thank your for the explanation!

PR #2976, opened 2016-04-19
change to config-renamed.properties

adnanb59:
- In line 11, is it raw.... or aw...?


slandelle:
A typo indeed. Could you please squash your commits into a single one?

adnanb59:
done

PR #2986, opened 2016-04-22
Recorder: Split password into keyStorePassword and keyManagerFactoryPassword

ggsjyoon:
in SslServerContext, ConfigKeys, RecorderConfiguration, and recorder-defaults.conf.


ggsjyoon:
I'm not sure why PaceSpec keeps timing out during CI, could it be a latency issue?

slandelle:
Thanks! I'm off atm, I'll merge later this week.

slandelle:
PaceSpec fails because some tests (Akka testkit based) expect some timings and Travis hosts frequently experience pauses/slowdowns.

slandelle:
I asked the original reporter for more information. I'd like to be sure that having different password isn't a very rare usage.

Then, this new property should be exposed in the GUI. And we'd have to define rules, eg when left blank, we should maybe fall back to the current behavior.

ggsjyoon:
I agree with you

slandelle:
This PR is broken and ships unrelated commits.
And we still don't know if this is useful for anyone.

Closing for now, will reopen if it makes sense.

Issue #2991, opened 2016-04-27
Exit signal is 0 even if assertions do not pass


michaelmcmillan:
Hi!

I am running into a problem with Gatling when ran on a Jenkins slave using Windows Server R2.

The exit code is 0 even if the assertions fail. I have attached the test case and the output. The last line in the output is there because I am running <code>echo Gatling exited with: %errorlevel%</code> after Gatling has executed.

````bash
Simulation intel.WSMPRead completed in 15 seconds
Parsing log file(s)...
Parsing log file(s) done
Generating reports...

================================================================================
---- Global Information --------------------------------------------------------
> request count                                       2000 (OK=99     KO=1901  )
> min response time                                   2797 (OK=2797   KO=3116  )
> max response time                                  15079 (OK=8616   KO=15079 )
> mean response time                                 12146 (OK=6180   KO=12457 )
> std deviation                                       3141 (OK=1362   KO=2886  )
> response time 50th percentile                      13108 (OK=6141   KO=13188 )
> response time 75th percentile                      13569 (OK=7291   KO=13602 )
> response time 95th percentile                      14485 (OK=8188   KO=14491 )
> response time 99th percentile                      14920 (OK=8523   KO=14930 )
> mean requests/sec                                    125 (OK=6.188  KO=118.812)
---- Response Time Distribution ------------------------------------------------
> t < 800 ms                                             0 (  0%)
> 800 ms < t < 1200 ms                                   0 (  0%)
> t > 1200 ms                                           99 (  5%)
> failed                                              1901 ( 95%)
---- Errors --------------------------------------------------------------------
> j.n.ConnectException: Connection refused: no further informati   1901 (100,0%)
on: localhost/127.0.0.1:8080
================================================================================

Reports generated in 1s.
Please open the following file: C:\Users\NTNU-stud\gatling-2.3.0\results\wsmpread-1461761125877\index.html
Global: percentage of successful requests is greater than 95 : false
Press any key to continue . . . 
Gatling exited with: 0
````

This is my test.

````scala
package intel

import io.gatling.core.Predef._
import io.gatling.http.Predef._
import scala.concurrent.duration._

class WSMPRead extends Simulation {

  val httpConf = http
    .baseURL("http://localhost:8080/INTEL_Server")
    .acceptHeader("text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
    .acceptLanguageHeader("en-US,en;q=0.5")
[...]

PR #2994, opened 2016-05-02
Forward the exit code from the Java process if it fails on Windows

michaelmcmillan:
Prior to this the process would always signal that it succeeded, making it hard to detect failed tests on Jenkins on top Windows.

Relevant issue: #2991


michaelmcmillan:
Does your test suite run on Windows? In that case I can add a test asserting the forwarding of exit signals are working.

slandelle:
Nope, it doesn't. That's the whole issue with Windows support: we're not Windows users.

slandelle:
@notdryft Do you have a Windows box at hand where you could give this change a try?

michaelmcmillan:
Nor am I, but I discovered the bug after being forced to run Jenkins on a Windows Server R2 environment. Kind of weird to think about that no one has run into this earlier ...

notdryft:
sure

michaelmcmillan:
Actually, I think my patch contains a bug. I added a comment with the <code>rem</code> function, I guess that <code>rem</code> will always succeed, thus ruining the error code check following it. Let me update the PR.

slandelle:
> run Jenkins on a Windows Server R2 environment

Sigh... Nothing shocks me anymore since I saw a big insurance company running Websphere AS on Windows NT ON PRODUCTION.

> Kind of weird to think about that no one has run into this earlier

I guess most ppl don't use the bundle but maven.

notdryft:
Looks good to me.

slandelle:
Thanks!

slandelle:
Thanks!

PR #3013, opened 2016-05-11
HTTP: Fixed SseHandler to also handle multiple data chunks correctly

mburger:
he guys, i was trying to use the sse support but without success because for example, this response which consists of 3 data chunks got processed wrong because the 3 data chunks represent one response through the sse channel but gatling processes the chunks one by one which i think is simply wrong
```
+--------+-------------------------------------------------+----------------+
|00000000| 35 0d 0a 64 61 74 61 3a 0d 0a 64 0d 0a 73 74 61 |5..data:..d..sta|
|00000010| 74 75 73 3a 20 52 45 41 44 59 0d 0a 32 0d 0a 0a |tus: READY..2...|
|00000020| 0a 0d 0a                                        |...             |
+--------+-------------------------------------------------+----------------+
```

this pull-request is only there to make you guys aware of the issue because this is just a very quick hack to fix the problem at hand so if you come up with something better feel free to decline it :)

### DIFF
--- a/gatling-bundle/src/universal/bin/gatling.bat
+++ b/gatling-bundle/src/universal/bin/gatling.bat
@@ -64,6 +64,8 @@ for %%i in ("%GATLING_HOME%\lib\*.jar") do call :addToPath "%%i"
 %JAVA% %COMPILER_OPTS% -cp %COMPILER_CLASSPATH% io.gatling.compiler.ZincCompiler -ccp %COMPILATION_CLASSPATH% %USER_ARGS%  2>NUL
 rem Run Gatling
 %JAVA% %JAVA_OPTS% -cp %GATLING_CLASSPATH% io.gatling.app.Gatling %USER_ARGS%
+if %errorlevel% neq 0 exit /b %errorlevel%
+rem The above line will forward any potential exit codes from Java if Gatling failed
 
 goto exit
 
--- a/gatling-bundle/src/universal/bin/gatling.sh
+++ b/gatling-bundle/src/universal/bin/gatling.sh
@@ -15,8 +15,6 @@
 # limitations under the License.
 #
 
-USER_ARGS="$@"
-
 OLDDIR=`pwd`
 BIN_DIR=`dirname $0`
 cd "${BIN_DIR}/.." && DEFAULT_GATLING_HOME=`pwd` && cd "${OLDDIR}"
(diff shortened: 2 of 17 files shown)


slandelle:
Hi @mburger 

If I get it right, your issue is that your message is spanned over multiple chunks, right?

I sadly think that the proper fix is going to be way more complicated than this.
We have no guarantee that some buffer will end up with "\n\n" at some point, all the more at high throughput. We should scan the buffer content for the last eol occurrence, and buffer all the bytes after that. You can see there the difficulty of chars vs bytes at SSE uses UTF-8 :(

Then, just looking for "\n\n" is wrong, as eol can be "\n", but also "\r\n", or "\r", see spec: https://www.w3.org/TR/2009/WD-eventsource-20091029/#event-stream-interpretation

mburger:
@slandelle 

exactly, according to https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.6.1 what happens is that gatling atm tries to parse each chunk but not that whole chunked-body like it would be required when using sse

and yes i totally agree with you that a proper fix would be more complicated then this quick hack :) the "\n\n" is just there because it works for what i need and it served to illustrate the problem at hand

would it be worth it if i update the pull request with a proper implementation over the next 1-2 weeks or do you want to take it from there ?

slandelle:
Being curious. How did you end up with messages spanned over multiple buffers? High throughput? Payloads larger than one MTU?

> would it be worth it if i update the pull request with a proper implementation over the next 1-2 weeks or do you want to take it from here ?

As you wish. Honestly, I'm swamped :)
But I can give you advice. For example, you could have a look at `Utf8ByteBuffersDecoder` for a hint on how to manually and effectively decode UTF-8 bytes.

If you can't tackle this yourself, providing some sample buffers (byte arrays) would surely help (we would need those for unit testing anyway).

mburger:
> Being curious. How did you end up with messages spanned over multiple buffers? High throughput? Payloads larger than one MTU?

Neither :)
This was just the first attempt using the sse support in gatling to test a simple Java App using the SseEmitter from the latest Spring Version running in a normal Tomcat 8, with one concurrent user from gatling and no additional load. The Messages themselves are also very small (around 76-300 bytes) with a normal MTU of 1500 and no fragmentation along the line.

> As you wish. Honestly, I'm swamped :)
But I can give you advice. For example, you could have a look at Utf8ByteBuffersDecoder for a hint on how to manually and effectively decode UTF-8 bytes.

Okidoki, then i'll see if i can cook something up in the next week if all goes well and advice is always welcome :)

I'll leave the PR open if thats ok for you and will update it accordingly

slandelle:
Closing this PR as the fix only works under some conditions (even if it's the most common one).
Opening issue #3134

Thanks a lot for your effort and your proposed fix!

slandelle:
@mburger would you be so kind as to give latest snapshot a try please?

mburger:
@slandelle just tested and it works like a charm :) thanks!

slandelle:
Great (and relieved)! Thanks for your feedback!

PR #3014, opened 2016-05-11
[new feature] Netflix/Eureka Integration

diegopacheco:
Hi Guys,

This is a new module. I did a very simple Netflix/Eureka integration to RESOLVE IPS via Netflix/Eureka REST API. There is a sample file in a directory gatling-eureka/sample/.

Cheers,
Diego Pacheco


slandelle:
Hi @diegopacheco,

Thanks for trying to contribute. But sadly, I think we won't merge this one. We chose to only integrate in Gatling main project mainstream features that will be widely used, and I don't think Eureka is there yet. Hype technologies come and goes.

Then, it seems the only thing you want is to discover your service's IP once on start-up. This would just be an utility class, not a protocol.

Once again, we love contributions, and we really appreciate your effort.

diegopacheco:
@slandelle 

Okay. Maybe consider create a contrib directory? 

Anyway the code is here is someone need it:
https://github.com/diegopacheco/gatling-eureka/

Cheers,
Diego Pacheco

PR #3045, opened 2016-06-09
Authorize on different server domain

welker90:
Can you add the ability to run a test with a user from a different domain that is trusted with the server's domain.  i.e. server is on domain "a", userID is in domain "b" and the 2 domains trust each other.
I've noticed this with ntlmauth for ntlmdomain.

Also with ntlm auth/http protocol - can you add the ability to have a feeder in http protocol so that you can have unique users "logging in" with ntlm auth?


slandelle:
It seems you've confused issues and pull requests (use for contributing).

Then please understand that NTLM and other MS entreprise authorization schemes are very hard to implement and test for open source projects as they require to set up a complex environment.

We can only investigate ourselves such topics through a sponsoring contract where a company would grant access to its environment.

Contributions are also very welcome.

Closing as this is not a proper pull request.

PR #3056, opened 2016-06-16
Add accumulated checks support for Websockets

BorisNaguet:
Hello & Bonjour,

On my current project we need to benchmark some scenario with http and websockets, but the current possibilities in gatling were not enought.

Let's say our server is a "long running task executor". You can send some task via a http POST, and it returns OK with a UUID when it's safely stored in a queue.
When the task is done, we send a status on an already opened websocket (identified by by client) with the task-id.

So on my scenario one User needs to 
- open a websocket 
-  POST lots tasks via http 
- adds a Check for each task with the returned UUID and status OK
- close websocket
We want to have timing for the http requests AND Checks 

Currently we can add a check without sending the message on websocket, but the problem was that we can't create a second check before the 1st one is matched (or it fails it with "Check didn't succeed by the time a new one was set up").

It was already discussed here:
https://groups.google.com/forum/#!topic/gatling/1kGyMC_0DyQ
http://stackoverflow.com/questions/26624448/issues-with-websocket-check-didnt-succeed-by-the-time-a-new-one-was-set-up

In the solution I propose, I wanted to keep the exact same behavior for existing scenarios, so I updated the DSL to add a new "accumulate" (I'm opened to better suggestion) option.
It was possible to keep the "blocking" option, but didn't make any sense to block and accumulate, so I forbid it at the DSL level.
Finally, I forced the "UntilCount(1)", since it would have been more difficult to keep track of counts for each Check (could be done later?).

Technically, I keep all AsynCheck instances in a Set (any better suggestion for a constantly evolving array? mutable.ListBuffer?). There's few risk of erasure, as they have a timestamp of creation property.
On a received message, I just loop on all accumulated checks, and on timeout I remove it from the Set and fail.

Functionally, it seems to work :) 
Of course, the size of the Set will evolve with your response time and finely tuned timeouts...

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/action/async/AsyncProtocolActor.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/action/async/AsyncProtocolActor.scala
@@ -15,18 +15,23 @@
  */
 package io.gatling.http.action.async
 
+import scala.collection.mutable
 import io.gatling.commons.stats.{ KO, OK, Status }
 import io.gatling.commons.util.Maps._
 import io.gatling.commons.util.TimeHelper._
 import io.gatling.core.action.Action
 import io.gatling.core.akka.BaseActor
-import io.gatling.core.check.CheckResult
+import io.gatling.core.check.{ Check, CheckResult }
 import io.gatling.core.session.Session
 import io.gatling.core.stats.StatsEngine
 import io.gatling.core.stats.message.ResponseTimings
 import io.gatling.http.check.async.AsyncCheck
 
 abstract class AsyncProtocolActor(statsEngine: StatsEngine) extends BaseActor {
+  /**
+   * These checks are not bound to an instance of AsyncTx
+   */
+  protected var accumulatedChecks = Set.empty[AsyncCheck]
 
   protected type NextTxBasedBehaviour = AsyncTx => this.Receive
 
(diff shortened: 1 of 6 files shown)


BorisNaguet:
I have a full test project test project with Gatling scenario and a mock self-contained server that I'll put on Github later ;)

slandelle:
@BorisNaguet Could you please DM me?

BorisNaguet:
Another issue I currently face in my scenario (I didn't realized it before) is that I can't define a "double check" on a websocket message (even without my PR)?
Maybe I missed it, is it possible to check that a single ws (json) message has:
_messId=${ID} && status='OK'_

On http, it's just 2 checks on a request.

Maybe with a headache-ing regex it's possible :)

slandelle:
WebSocket API will be completely rebooted in Gatling 3.
Effort should go there: #3071.

Please join the discussion in there.

BorisNaguet:
Any news on a "how to contribute", for an individual or a company?
I haven't seen any post on the blog about that or any.

slandelle:
We're in the process of adding individual and corporate CLA.
Then (nothing's decided yet), regarding contributing fixes on existing code can easily be done. We take care of integrating in the current public branch and porting in the next major or minor release too.

But contributing to the future code is another matter. We only plan to grant access to:
* very active community members
* customers who went with an extended support plan to get the sources, not only the milestones binaries

If you or your company are interested in discussing about this, please get in touch at contact@gatling.io.

PR #3057, opened 2016-06-20
Save system COMPILATION_CLASSPATH variable for jar usage ability

greyfenrir:
I'm writing automation wrapper around Gatling and need to use external java class in Gatling scala script. 
I can tell java where this class is located, but I can't tell it zinc compiler due to $COMPILATION_CLASSPATH takes files only from $GATLING_HOME/lib. 
It would be great if I could add my *jar to zinc libraries. For this it's enough to use same environment variable.


slandelle:
Not a supported usage.

PR #3062, opened 2016-06-22
HAR import: Correctly parse the date when the millisecond field conta…

dmfreemon:
…ins more than three digits


slandelle:
Oh right. I guess this code comes from early days when we were supporting JDK6 and java.time didn't exist. Thanks!

slandelle:
Cherry-picked as fbbbde2131bc7740b9c1390614935823a6c5a7a0

PR #3063, opened 2016-06-23
Doc: Realtime monitoring with InfluxDB & Grafana

aidylewis:


notdryft:
We have a meeting tonight so are kind of swamped right now... we will review this tomorrow morning !

slandelle:
My main issue with this PR is that it completely removes any mention to Graphite.
It lets Graphite users think that we don't support it, while it's actually the primary support and we only talk to InfluxDB over the Graphite protocol.
Then, some organizations have been using Graphite successfully for ages and are not willing to drop it for the InfluxDB/Grafana combo. Some even might be using Carbon (the Graphite DB) with a Grafana dashboard.

The documentation improvements you suggest are great, but I'd rather not drop Graphite support.

aidylewis:
OK. I'll add the Graphite tools.

slandelle:
Thanks!

aidylewis:
I don't know enough about the Graphite suite of tools to go around hacking. How about keep what is, but add a subpage: "InfluxDB and Grafana" (or just append it to the bottom)?

aidylewis:
Added Graphite config.  [synthesize](https://github.com/obfuscurity/synthesize) comes with collectd, so kept the collectd config.

src/sphinx/realtime_monitoring/index.rst
```diff
@@ -4,102 +4,115 @@
...
+$ sudo service influxdb start
 
-Configuration
-=============
+Verification
+------------
 
-Gatling needs to be configured to send metrics.
-In addition the tool that recieves and processes the data needs to be configured. For example one of Graphite, InfluxDB or other timeseries database that supports the Graphite protocol.
+Run the `example simulation <https://github.com/bbc/example-gatling-sbt#run-a-simulation>`_, and
```
slandelle: Sorry, but we'd keep all Gatling related resources in our own organization.

aidylewis:
Ok. I'll send a fix by the end of today.

aidylewis:
What do you think about adding this sim: https://github.com/bbc/example-gatling-sbt/blob/master/src/test/scala/kraftwerk/ComputerWorld.scala (without print statements) to https://github.com/gatling/gatling-sbt-plugin-demo ?

aidylewis:
Also, I have a gatling-grafana-template. What do you want to do about this? 
https://github.com/aidylewis/gatling-grafana-template

slandelle:
> What do you think about adding this sim: https://github.com/bbc/example-gatling-sbt/blob/master/src/test/scala/kraftwerk/ComputerWorld.scala (without print statements) to https://github.com/gatling/gatling-sbt-plugin-demo ?

Sure

> Also, I have a gatling-grafana-template. What do you want to do about this? 
https://github.com/aidylewis/gatling-grafana-template

Move to `src/sphinx/realtime_monitoring/code` ?

aidylewis:
I think that's it from me!!!

slandelle:
Cool, thanks!

slandelle:
I'll add some more information about FrontLine.

slandelle:
Cherry-pick as 90b4bd64215c6ef61c4b572939a52a4c13fbbb9d

Thanks a lot!

PR #3079, opened 2016-07-11
Reports: Eliminate the possibility of overflow

qzhjason:
When test has very large sum number.

### DIFF
--- a/gatling-charts/src/main/scala/io/gatling/charts/stats/buffers/GeneralStatsBuffers.scala
+++ b/gatling-charts/src/main/scala/io/gatling/charts/stats/buffers/GeneralStatsBuffers.scala
@@ -84,8 +84,8 @@ private[stats] class GeneralStatsBuffer(durationInSec: Long) {
 
     } else {
       val count = digest.size
-      val mean = (sum / count).toInt
-      val stdDev = math.sqrt((sumOfSquares / count - mean * mean).toInt
+      val mean = (sum / count).toDouble
+      val stdDev = math.sqrt(sumOfSquares / count - mean * mean).toInt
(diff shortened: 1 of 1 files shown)


slandelle:
@qzhjason Nice catch, thanks!

Then, your fix doesn't compile (unmatched parens) and it would probably be better to have a double mean too.
Can you please fix and rebase your commit?

qzhjason:
I'll try to fix it on my local PC.

gatling-charts/src/main/scala/io/gatling/charts/stats/buffers/GeneralStatsBuffers.scala
```diff
@@ -84,8 +84,8 @@ private[stats] class GeneralStatsBuffer(durationInSec: Long) {
 
     } else {
       val count = digest.size
-      val mean = (sum / count).toInt
-      val stdDev = math.sqrt((sumOfSquares - (sum * sum) / count) / count).toInt
+      val mean = (sum / count).toDouble
```
slandelle: This doesn't do much. You want to turn sum into a double, then divide.

gatling-charts/src/main/scala/io/gatling/charts/stats/buffers/GeneralStatsBuffers.scala
```diff
@@ -84,8 +84,8 @@ private[stats] class GeneralStatsBuffer(durationInSec: Long) {
...
+      val stdDev = math.sqrt(sumOfSquares / count - mean * mean).toInt
```
slandelle: sumOfSquares.toDouble

slandelle:
Could you please squash your commits into a single one?

PR #3082, opened 2016-07-17
Core: Put user counters at the end to see them without scrolling when there are a lot of different requests

henri-tremblay:


slandelle:
Thanks, will give it a try this week.

slandelle:
Mmm, actually, when there's lots of requests, at some point you'd have to scroll anyway, and your change makes it harder to see how users are advancing.
IMO, it makes more sense to leave as is.

@notdryft WDYT?

slandelle:
Me stupid, missing your point.

slandelle:
Cherry-picked, thanks man!

henri-tremblay:
Thanks!

PR #3092, opened 2016-07-27
Better calculate a mean with 'double' than with 'int'

nremond:
Minor, but avoids the lost of precision.


nremond:
Also, is there a reason for handling the (double->int) conversion without rounding first ?

slandelle:
Thanks man!

Issue #3097, opened 2016-08-03
Allow graphite configuration to also use environment variables and not only gatling.conf


camiloribeiro:
I am using gatling as engine for running performance testing and monitoring in a fatjar that we spread around several datacenters, so it is hard (and unnecessary) to build different fatjars because I need two different configurations, one for each data center for instance. 

The feature would be to keep the current behaviour and allow users to use env variables to set values. If the environment variable is not there, it fallback to its current behaviour and uses gatling.config as usual.

the configuration I am asking to be changed is, with its respective environment variable suggested names:

```
 graphite {
      light = false                         # GRAPHITE_LIGHT
      host = "metrics.tvoli.com"            # GRAPHITE_HOST
      port = 2003                           # GRAPHITE_PORT
      protocol = "tcp"                      # GRAPHITE_PROTOCOL
      rootPathPrefix = "production"         # GRAPHITE_ROOTPATHPREFIX
      bufferSize = 8192                     # GRAPHITE_BUFFERSIZE
      writeInterval = 5                     # GRAPHITE_WRITEINTERVAL
    }
```

I know it sounds very specific, but this would not affect the current api and would not change the behaviour we have today.

Let me know if you need any extra data. A PR is coming with this issue soon.

slandelle:
PR reviewed, but I had to reject it.
Please use Lightbend config's built-in features.

slandelle:
So, all good with plain config lib?

PR #3098, opened 2016-08-03
Adding environment variable support to graphite configuration. Feature #3097

camiloribeiro:
Using a clean way to implement the feature request with the suggested names. I could not find a nice way to test it :( 

Please let me know if anything else is needed of if it is too specific. Thanks.


slandelle:
Hi.

Thanks for your effort, I disagree with your implementation.

First, it would be inconsistent to only have Graphite related config resolved against environment variables and not the rest of it.

Then, we use Lightbend's [config library](https://github.com/typesafehub/config) that lets you override props defined in the config files with System properties (ex: `-Dgatling.data.graphite.light=true`).
Could you please check if it's also enabled by default for environment variables too?
If it doesn't, you could create a `Config` from environment (`ConfigFactory.systemEnvironment`) and add that to the fallback chain (see `GatlingConfiguration.loadForTest`).
In this case, I'm looking forward the next version of your pull request :)

camiloribeiro:
Thanks for letting me know! I will check and let you know. Thanks.

PR #3124, opened 2016-09-01
fix GATLING_HOME detection when $0 is symlink

ryenus:
When gatling is launched via a symlink, e.g. `/usr/local/bin/gatling -> /opt/gatling/bin/gatling.sh` 
It fails with the following error:
> GATLING_HOME is set to /usr/local
Error: Could not find or load main class io.gatling.app.Gatling

BTW, for better cross platform compatibility, `readlink` options  `-e` or `-f` are not used, as that's GNU specific, not available in `/usr/bin/readlink` on Mac. the effect is the resolved `GATLING_HOME` could be still a symlink, but feature wise it works fine.

### DIFF
--- a/gatling-bundle/src/universal/bin/gatling.sh
+++ b/gatling-bundle/src/universal/bin/gatling.sh
@@ -15,7 +15,25 @@
 # limitations under the License.
 #
 
-DEFAULT_GATLING_HOME=$(cd "$(dirname "$(readlink -n "$0")")/.."; pwd)
+
+## resolve links - $0 may be a link to Maven's home
+PRG="$0"
+
+# need this for relative symlinks
+while [ -h "$PRG" ] ; do
+  ls=`ls -ld "$PRG"`
+  link=`expr "$ls" : '.*-> \(.*\)$'`
+  if expr "$link" : '/.*' > /dev/null; then
+    PRG="$link"
+  else
+    PRG="`dirname "$PRG"`/$link"
+  fi
+done
+
+DEFAULT_GATLING_HOME=`dirname "$PRG"`/..
+
+# make it fully qualified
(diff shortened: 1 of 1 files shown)


slandelle:
@ryenus Thanks!
@notdryft Please review

notdryft:
This modification would make the script fail for the most common case, which is running `./bin/gatling.sh` from the root directory of a fresh downloaded bundle:

```bash
$ pwd
/Users/notdryft/Playground
$ tree
.
└── bin
    └── test.sh

1 directory, 1 file
$ cat bin/test.sh
#!/bin/sh

#OLDDIR=`pwd`
#BIN_DIR=`dirname $0`
#cd "${BIN_DIR}/.." && DEFAULT_GATLING_HOME=`pwd` && cd "${OLDDIR}"
DEFAULT_GATLING_HOME=$(cd "$(dirname "$(readlink -n "$0")")/.."; pwd)

GATLING_HOME="${GATLING_HOME:=${DEFAULT_GATLING_HOME}}"

echo "GATLING_HOME is set to ${GATLING_HOME}"
$ ./bin/test.sh
$ ln -s /Users/notdryft/Playground/bin/test.sh /usr/local/bin/test.sh
$ test.sh
GATLING_HOME is set to /Users/notdryft/Playground
```

So, maybe detect the nature of the link and use readlink or the former method depending of the result?

ryenus:
@notdryft is right, sadly `readlink -n ./bin/test.sh` never print anything, be it `/usr/bin/readlink` on Mac, or GNU `readlink` from coreutils.

Maybe the solution is to [resort to `perl`][1] first, then fallback to gnu `readlink`?

[1]: http://stackoverflow.com/a/24572274/537554

slandelle:
@notdryft please review again

slandelle:
I'm sorry, but I'd rather not take the risk of breaking something that works for the sake of supporting a rare and non standard usage.

PR #3128, opened 2016-09-02
Improve mvn plugin doc (multiple simulation & incl/excl)

ronanquillevere:
Add section dedicated to the use of runMultipleSimulations with includes and excludes syntax examples


slandelle:
Great, thanks!
Could you please r[ebase and squash](http://stackoverflow.com/questions/3103589/how-can-i-easily-fixup-a-past-commit) your commits into a single one?

PR #3130, opened 2016-09-05
Doc: run multiple simulations with includes and excludes with maven plugin

ronanquillevere:
Improve documentation with example of includes and excludes configuration options.


slandelle:
cherry-picked

slandelle:
Thanks!

PR #3147, opened 2016-09-19
Use JAVA_HOME on Windows when running the recorder

z0li:
JAVA_HOME is hanadled the same way as in gatling.bat


slandelle:
Nice catch, thanks!

slandelle:
cherry-pick as abee52809ff32b36895759e18e11669171a96a58

Issue #3150, opened 2016-09-21
App: JAVA_OPTIONS should be prepended


slandelle:
Instead of appended so user defined props have precedence (Hotspot gives precedence to rightmost values).

slandelle:
Closed by e9256ccba655b3ea7911fe965a6e9c9139562020

PR #3151, opened 2016-09-22
User defined JAVA_OPTS take precedence

reikje:
Fix for https://github.com/gatling/gatling/issues/3150

### DIFF
--- a/gatling-bundle/src/universal/bin/gatling.sh
+++ b/gatling-bundle/src/universal/bin/gatling.sh
@@ -26,12 +26,18 @@ export GATLING_HOME GATLING_CONF
 
 echo "GATLING_HOME is set to ${GATLING_HOME}"
 
-JAVA_OPTS="${JAVA_OPTS} -server"
-JAVA_OPTS="${JAVA_OPTS} -Xmx1G"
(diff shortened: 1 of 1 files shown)


slandelle:
Thanks, but I wonder about that.
The behavior we were trying to achieve was to amend the user defined props. But in order to do so, we should have place Gatling props on the left of the user defined ones, not on the left.
Your change is a breaking one: users providing options that don't "conflict" with the Gatling ones would lose the laters.
We should stick with expected behavior atm.

Then, we also need the recorder scripts and the Windows ones to get fixed too, behavior should be the same.

WDYT?

reikje:
Hmm isn't it dangerous to mix them and rely on a specific order? Will it work on all JVM's? What about illegal combinations, i.e. user specifying `-Xms2G` and gatling specifying `-Xmx1G`. Maybe there could be an include file like `cassandra.in.sh` (Cassandra) or `setenv.sh` (Tomcat) uses? So the user would have to edit this file instead of sending custom arguments via the `JAVA_OPTS` environment variable.

ps. will update other start scripts after we have sorted Linux :)

slandelle:
> Will it work on all JVM's?

OpenJDK, Hotspot and J9: yes. Can't say for Zing, but I seriously doubt anyone would pay for it to use it for Gatling ;)

We can go with the changes you suggest for a future version but not in a fix.

slandelle:
I went with user defined JAVA_OPTS taking precedence but overloading instead of replacing, so that users can "patch" Gatling options.

PR #3153, opened 2016-09-22
Core: TryMax should support an Expression[Int]

GabrielPlassard:
Hi,
unless there is a reason not to allow it, this should make it possible to use Expression[Int] for tryMax. :-)
With a test simulation : https://github.com/GabrielPlassard/gatlingTest/blob/master/src/test/scala/gatlingtest/TryMaxSimulation.scala

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/action/TryMax.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/action/TryMax.scala
@@ -19,13 +19,13 @@ import scala.concurrent.duration._
 
 import io.gatling.commons.stats.KO
 import io.gatling.commons.validation._
-import io.gatling.core.session.{ Session, TryMaxBlock }
+import io.gatling.core.session.{ Expression, Session, TryMaxBlock }
 import io.gatling.core.stats.StatsEngine
 import io.gatling.core.util.NameGen
 
 import akka.actor.ActorSystem
 
-class TryMax(times: Int, counterName: String, statsEngine: StatsEngine, next: Action) extends Action with NameGen {
+class TryMax(times: Expression[Int], counterName: String, statsEngine: StatsEngine, next: Action) extends Action with NameGen {
 
   override val name = genName("tryMax")
 
@@ -40,7 +40,7 @@ class TryMax(times: Int, counterName: String, statsEngine: StatsEngine, next: Ac
 }
 
 class InnerTryMax(
-    times:       Int,
+    times:       Expression[Int],
     loopNext:    Action,
     counterName: String,
     system:      ActorSystem,
@@ -61,11 +61,18 @@ class InnerTryMax(
     case _                           => false
   }
 
-  private def maxNotReached(session: Session): Boolean = session(counterName).validate[Int] match {
-    case Success(i) => i < times
-    case Failure(message) =>
-      logger.error(s"Condition evaluation for tryMax $counterName crashed with message '$message', exiting tryMax")
-      false
+  private def maxNotReached(session: Session): Boolean = {
+    val validationResult = for {
+      counter <- session(counterName).validate[Int]
+      max <- times(session)
+    } yield counter < max
+
+    validationResult match {
+      case Success(maxNotReached) => maxNotReached
+      case Failure(message) =>
+        logger.error(s"Condition evaluation for tryMax $counterName crashed with message '$message', exiting tryMax")
+        false
+    }
   }
 
   private def continue(session: Session): Boolean = blockFailed(session) && maxNotReached(session)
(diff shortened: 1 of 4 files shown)


gatling-core/src/main/scala/io/gatling/core/action/TryMax.scala
```diff
@@ -16,16 +16,14 @@
...
 
 import scala.concurrent.duration._
-
 import io.gatling.commons.stats.KO
 import io.gatling.commons.validation._
-import io.gatling.core.session.{ Session, TryMaxBlock }
+import io.gatling.core.session.{ Expression, Session, TryMaxBlock }
 import io.gatling.core.stats.StatsEngine
 import io.gatling.core.util.NameGen
-
```
slandelle: Please respect of formatting rules

gatling-core/src/main/scala/io/gatling/core/action/TryMax.scala
```diff
@@ -61,11 +59,18 @@ class InnerTryMax(
...
+      counter <- session(counterName).validate[Int]
+      max <- times(session)
+    } yield (counter, max)
+
+    validationResult match {
+      case Success((counter, max)) => counter < max
+      case Failure(message) =>
+        logger.error(s"Condition evaluation for tryMax $counterName crashed with message '$message', exiting tryMax")
+        false
+    }
```
slandelle: Please use one single for-comprehension, without this intermediate `validationResult` and this additional pattern matching.

gatling-core/src/main/scala/io/gatling/core/structure/Errors.scala
```diff
@@ -17,14 +17,14 @@ package io.gatling.core.structure
...
 import io.gatling.core.action.builder.{ ExitHereIfFailedBuilder, TryMaxBuilder }
+import io.gatling.core.session.Expression
 
 trait Errors[B] extends Execs[B] {
 
-  def exitBlockOnFail(chain: ChainBuilder): B = tryMax(1)(chain)
-  def tryMax(times: Int, counter: String = UUID.randomUUID.toString)(chain: ChainBuilder): B = {
-
-    require(times >= 1, "Can't set up a max try <= 1")
+  def exitBlockOnFail(chain: ChainBuilder): B = tryMax(_ => Success(1))(chain)
```
slandelle: Use `expressionSuccess` from the `session` package object so it doesn't instantiate a `Success` object every time the function is applied.

slandelle:
Thanks :)

Please address comments.

gatling-core/src/main/scala/io/gatling/core/action/TryMax.scala
GabrielPlassard: Not sure how I should get rid of the pattern matching, recover + get ?

slandelle: You're right, forget it

GabrielPlassard:
Should be fixed, although I was not sure what I should really put in HttpCompileTest

gatling-core/src/main/scala/io/gatling/core/action/TryMax.scala
```diff
@@ -61,11 +61,18 @@ class InnerTryMax(
...
-  private def maxNotReached(session: Session): Boolean = session(counterName).validate[Int] match {
-    case Success(i) => i < times
-    case Failure(message) =>
-      logger.error(s"Condition evaluation for tryMax $counterName crashed with message '$message', exiting tryMax")
-      false
+  private def maxNotReached(session: Session): Boolean = {
+    val validationResult = for {
+      counter <- session(counterName).validate[Int]
+      max <- times(session)
+    } yield (counter, max)
```
slandelle: Actually, you can directly yield `counter < max`

gatling-core/src/main/scala/io/gatling/core/structure/Errors.scala
```diff
@@ -18,13 +18,13 @@ package io.gatling.core.structure
 import java.util.UUID
 
 import io.gatling.core.action.builder.{ ExitHereIfFailedBuilder, TryMaxBuilder }
+import io.gatling.core.session._
+import io.gatling.core.session.Expression
```
slandelle: useless import

GabrielPlassard:
Done!

slandelle:
Perfect, thanks!
Will cherry-pick tomorrow

slandelle:
cherry-picked as b54959d3d4e013ec90f778b78a8e18428757af7b

Thanks!

PR #3173, opened 2016-10-29
Upgrade sbt 0.13.13

pdalpra:
This is the most straightforward migration path that I could come up with, a local AutoPlugin could help a bit by getting rid of duplication of:

``` scala
.enablePlugins(SonatypeReleasePlugin)
.settings(gatlingModuleSettings: _*)
```

There's probably some other tricks that could be applied to make it simpler, such as overriding the project name to avoid renaming `app` to `gatling-app` to keep the same module/artifact name.
I'll have a further look in the afternoon.


slandelle:
Awesome. We have a big release train coming, so we'll merge soon after.
Thanks a lot!

pdalpra:
I tried to override the project name, without any luck, maybe I'm missing something or that just doesn't work :/

notdryft:
From what I saw, `gatlingModule`can be kept. However, `project.in` can only be done on a `val`:

```bash
error: project must be directly assigned to a val, such as `val x = project`.
```

Still, I was able to use the `Project(id, file(id))`construct inside a `def`.

pdalpra:
@notdryft Yes, the `project` macro is a bit too restrictive on that, as it uses the name of the `val` to set the project name.

> Still, I was able to use the Project(id, file(id))construct inside a def.

On the future availability of `Project(id: String, base: File)`, it seems it's on it's way out too: https://github.com/sbt/sbt/pull/2524

notdryft:
Oh boy >.>

notdryft:
ScalaJs kept the `Project(...) construct`: https://github.com/scala-js/scala-js/commit/aad1c895842b578d910a463b25b661d721613b2a :<

notdryft:
> I'm going to remove Project(..., settings = ) constructor soon, so that'll require those users to change the build file anyway.

As this is not the one where are using, maybe we can get away for now with the simplest form!

pdalpra:
@notdryft I'll update my PR accordingly, thx for the info !

PR #3206, opened 2016-12-05
Fix cheatsheet for conditionals

suzannehamilton:
I was writing a scenario with some conditionals, and noticed that the [cheatsheet](http://gatling.io/#/cheat-sheet/2.2.3) docs for conditionals weren't quite right:

* No entry for `doIfEquals`
* The parameters for `doIf` in the cheatsheet are `(actual, expected)`, but it should be a single expression parameter (or something which can be implicitly converted to one)

I've fixed those in this pull request. I've tried to stick with the conventions in the existing cheatsheet entries, which give a quick summary of the usage without going into all the possibilities provided by implicit conversion.

### DIFF
--- a/src/sphinx/cheat-sheet.html
+++ b/src/sphinx/cheat-sheet.html
@@ -43,10 +43,11 @@
           <div class="function-details">
             Executes a chain when a condition is satisfied
             <dl>
-              <dt>(actual, expected){chain}</dt>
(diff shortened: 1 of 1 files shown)


slandelle:
@suzannehamilton Thanks a lot!

cherry-picked as fdf5a8f06a9fa0fe59194f949c96a2ecf7b1dee7

PR #3208, opened 2016-12-07
Changing println statements to logger.info

svetaa:
Hi there,

I've had to go through unnecessary workarounds, not being able to control the logging. Is it possible to switch these println statements to logger.info?

Thanks,

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/Runner.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/Runner.scala
@@ -29,6 +29,7 @@ import io.gatling.core.controller.throttle.Throttler
 import io.gatling.core.scenario.{ Scenario, SimulationParams }
 import io.gatling.core.stats.{ DataWritersStatsEngine, StatsEngine }
 import io.gatling.core.stats.writer.RunMessage
+
 import akka.actor.ActorSystem
 import akka.pattern.ask
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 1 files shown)


slandelle:
What's your problem exactly?

slandelle:
Those logs are definitively useful for newcomers as they will do many mistakes first and need to see something happening so they know things are making progress, all the more as the Scala compiling phase can take a while. So we definitively need to be able to display them by default.

They use println because Gatling's logback is configured with WARN level enabled.
It would neither make sense to turn them in WARN or ERROR logback logs.
Nor would is make sense to log them at INFO level and enable INFO level globally.

Then, we don't know what people have been doing with those logs so far. Some people might use them, can't tell.

In the end, I know what we could lose with the change you suggest, but not what we would gain.
Please provide some details on how it would benefit most users.

svetaa:
I see. Thank you for the detailed explanation. I have one use case with a gatling project called using a custom shell script meant for interactive use. Within the context of the script, referencing simulations in console logs does not make sense and I needed to restrict script output to a minimal, predefined format. I was able to turn off WARN level logs in logback to achieve what I needed, except these two simulation messages.

slandelle:
You still haven't explained the full story. What is that interactive shell doing?
But well, I guess I'd have to [ask 5 times](https://en.wikipedia.org/wiki/5_Whys)... :)

Anyway, I'm not inclined to turn those println to slf4j logs, for the mentioned reasons.

What would work, thought, would be only enable those when the `mute` option is off (default).
If that would work for you, would you like to provide another PR?

PR #3231, opened 2017-02-06
Doc: Capture all Graphite data

aidylewis:
Simplified InfluxDB templates which now capture all Graphite data from the sample simulation and placing them into two measurements; *gatling* and *gatling.users*.


slandelle:
cherry-picked, thanks!

PR #3257, opened 2017-03-08
App: call LoggerContext.stop() at the end of simulation, close #3255

### LINKED ISSUE #3255: App: Some error message is dropped at tail of logback log file
Logback buffer is not flshed on finising simulation.
Some error message is dropped at tail of ``target/logs/test.log``.
(I tested using SBT.)

I think ``loggerContext.stop()`` should be called at the end of run.

t-takatsuru:
Fixed to call ``LoggerContext.stop()`` at the end of run.
Statements in ``start(...)`` method are surrounded by try-finally, and log buffer will be flushed even if simulation is stopped with Ctrl-C.
This will work both run in SBT and zipped bundle.

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/Gatling.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/Gatling.scala
@@ -23,7 +23,9 @@ import io.gatling.app.cli.ArgsParser
 import io.gatling.core.config.GatlingConfiguration
 
 import akka.actor.ActorSystem
+import ch.qos.logback.classic.LoggerContext
 import com.typesafe.scalalogging.StrictLogging
+import org.slf4j.LoggerFactory
 
 /**
  * Object containing entry point of application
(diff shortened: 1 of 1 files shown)


slandelle:
@t-takatsuru Could you please point me to where calling `LoggerContext#stop` triggers appender flush? I've had a very quick look at logback's sources and couldn't find it.
Thanks in advance!

t-takatsuru:
FileAppender uses BufferedOutputStream as its buffer.
(see logback-core/src/main/java/ch/qos/logback/core/FileAppender.java:240)

When ``LoggerContext.stop()`` is called, ``OutputStream.close()`` is called in following call stack:
```
LoggerContext.stop()  (ch.qos.logback.classic)
LoggerContext.reset()  (ch.qos.logback.classic)
Logger.recursiveReset()  (ch.qos.logback.classic)
Logger.detachAndStopAllAppenders()  (ch.qos.logback.classic)
AppenderAttachableImpl.detachAndStopAllAppenders()  (ch.qos.logback.core.spi)
LifeCycle.stop()  (ch.qos.logback.core.spi)
FileAppender.stop()  (ch.qos.logback.core)
OutputStreamAppender.stop()  (ch.qos.logback.core)
OutputStreamAppender.closeOutputStream()  (ch.qos.logback.core)
OuputStrem.close() (java.io)
```

Almost appenders write logs immediately.
Without immediateFlush flag, FileAppender does not write immediately, and flushes BufferedOutputStream on ``LifeCycle.stop()`` at one time.

slandelle:
@t-takatsuru Thanks!

This implementation assumes the `Gatling` class owns the `LoggerContext`.
I want to check that this is always the case, and that we don't sometimes have a caller owner, such as our maven and sbt plugins.

slandelle:
cherry-picked.

Thanks a lot!

slandelle:
@t-takatsuru FYI, your code was broken :)
I failed to notice it as it did compile, even though your code seemed a bit strange, so my bad.

```scala
val loggerContext: LoggerContext = LoggerFactory.getILoggerFactory.asInstanceOf
loggerContext.stop()
```

As you didn't pass a type parameter to `asInstanceOf`, the Scala compiler was actually using `Nothing` instead of `LoggerContext`. This caused a `ClassCastException` when trying to assign to the left reference which is of type `LoggerContext`.
Good Scala type inference puzzler :)

I turned it into:
```scala
 LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext].stop()
```

t-takatsuru:
Thanks for your notice :)
My study of Scala is not enough...

Issue #3256, opened 2017-03-07
Recorder: Content-Type header value is empty in simulation generated from HAR.


t-takatsuru:
Content-Type header is empty in simulation generated from HAR.

protocol part of ``RecordedSimulation.scala`` is following:
```scala
val httpProtocol = http
    .baseURL("https://xxx.xxx.xxx.xxx")
    .contentTypeHeader("")
```

This simulation produces HTTP request with empty value Content-Type header in every interaction without overriding with ``.header(...)`` method.
It is problem on interpreting request in target web server.

Attachment HAR file to reproduce that.
[sample.har.txt](https://github.com/gatling/gatling/files/823599/sample.har.txt)

slandelle:
Hi again, and thanks again :)

Same question as [here](https://github.com/gatling/gatling/issues/3255#issuecomment-284657628).

PR #3258, opened 2017-03-08
removed unnecessary Content-Type default. #3256

t-takatsuru:
removed unecessary ``.contentTypeHeader("")`` and generates header method for each interaction properly.


slandelle:
@t-takatsuru Actually, the issue lays way upstream, [here](https://github.com/gatling/gatling/blob/master/gatling-recorder/src/main/scala/io/gatling/recorder/har/HarReader.scala#L91).

I don't know if things were different in older Chrome versions, but it seems `postData.mimeType` is now always defined, yet possibly empty.
I've fixed it and will push the fix soon.
Thanks a lot for your effort! Please keep those issue reports and PR coming!

PR #3260, opened 2017-03-08
Recorder: Decode form parameters that are url encoded when importing HAR files, close #3259

### LINKED ISSUE #3259: Gatling recorder HAR imports x-www-form-urlencoded form parameters in encoded form
I have a simple form that has the enctype "application/x-www-form-urlencoded". When my form parameters contain characters that should be encoded in the URL the Gatling recorder will add these to the scenario in encoded form when using the HAR import function (resulting in double ending when the request is executed). 
When using the recorder in Proxy mode the parameters are properly decoded.

Reproducer HAR-file:
https://github.com/timve/gatling/blob/Fix_HAR_urlencoded_form_parameters/gatling-recorder/src/test/resources/har/form-encoding.har

This should result in a line:
`.formParam("param", "H@llo"))`

instead of:
`.formParam("param", "H%40llo"))`

timve:


gatling-recorder/src/main/scala/io/gatling/recorder/har/HarReader.scala
```diff
@@ -58,17 +60,26 @@ private[recorder] object HarReader {
...
     val requestHeaders = buildRequestHeaders(entry)
 
     // NetExport doesn't copy post params to text field
     val requestBody = entry.request.postData.flatMap { postData =>
-      if (postData.params.nonEmpty)
-        Some(buildContent(postData.params))
-      else
+      if (postData.params.nonEmpty) {
+        val requestContentType = requestHeaders.get(ContentType)
+        val isUrlEncoded = requestContentType.exists(_.contains(ApplicationFormUrlEncoded))
```
slandelle: Why `contains`? Why not `equalsIgnoreCase`?

gatling-recorder/src/test/resources/har/form-encoding.har
```diff
@@ -0,0 +1,231 @@
...
+          "receive": 0.5050000036135267,
+          "ssl": -1
+        },
+        "serverIPAddress": "127.0.0.1",
+        "connection": "75288",
+        "pageref": "page_10"
+      }
+    ]
+  }
+}
```
slandelle: missing line feed so git is happy

slandelle:
Thanks @timve ! Please address comments :)

gatling-recorder/src/main/scala/io/gatling/recorder/har/HarReader.scala
timve: copy-paste from `ScenarioElement`...
I think `equalsIgnoreCase` is not the right method to use here because the HTTP Content-Type header contains a media-type value (https://tools.ietf.org/html/rfc7231#section-3.1.1.5), which can contain parameters (https://tools.ietf.org/html/rfc7231#section-3.1.1.1)

gatling-recorder/src/test/resources/har/form-encoding.har
timve: added the line feed

gatling-recorder/src/main/scala/io/gatling/recorder/har/HarReader.scala
slandelle: Good point

slandelle:
@timve Could you please squash your commits so I can easily cherry-pick?

slandelle:
Madre de dios! rebase fail :(

timve:
yep, me being a git noob, trying to fix the mess, if I fail I will make a new PR for you ;-)

timve:
Ok, that seems to have done the trick

slandelle:
cherry-picked

Thanks a bunch!

PR #3262, opened 2017-03-11
JMS: support custom JNDI properties

pwielgolaski:
In some cases you could need to specific JNDI properties for JMS, at the moment you need can not do it using builder, you need to create connection factory by yourself.
This PR fix it

### DIFF
--- a/gatling-jms/src/main/scala/io/gatling/jms/jndi/JmsJndiConnectionFactoryBuilder.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/jndi/JmsJndiConnectionFactoryBuilder.scala
@@ -19,6 +19,8 @@ import java.util.{ Hashtable => JHashtable }
 import javax.jms.ConnectionFactory
 import javax.naming.{ Context, InitialContext }
 
+import scala.collection.JavaConverters._
+
 import io.gatling.commons.model.Credentials
 
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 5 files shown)


gatling-jms/src/main/scala/io/gatling/jms/jndi/JmsJndiConnectionFactoryBuilder.scala
```diff
@@ -64,6 +68,10 @@ case class JmsJndiConnectionFactoryBuilder(
       properties.put(Context.SECURITY_CREDENTIALS, credentials.password)
     }
 
+    jndiProperties.foreach { case (k, v) =>
+      properties.put(k, v)
+    }
```
slandelle: Add in imports `import scala.collection.JavaConverters._`

then replace with `properties.putAll(jndiProperties.asJava)`

slandelle:
@pwielgolaski Thanks!
Could you please address comment.
Also, for completeness, could you please add a call to the new DSL method in `JmsCompileTest` and doc is `jms.rst`?

pwielgolaski:
Thank you for your comments, I applied them and it should be good to merge.

gatling-jms/src/main/scala/io/gatling/jms/jndi/JmsJndiConnectionFactoryBuilder.scala
```diff
@@ -37,20 +39,24 @@ case class JmsJndiConnectionFactoryBuilderUrlStep(connectionFactoryName: String)
...
     connectionFactoryName: String,
     url:                   String,
-    credentials:           Option[Credentials] = None
+    credentials:           Option[Credentials] = None,
+    properties:            Map[String, String] = Map.empty
 ) {
 
   def credentials(user: String, password: String) = copy(credentials = Some(Credentials(user, password)))
 
+  def property(key: String, value: String) = copy (properties = properties + (key -> value))
```
slandelle: I missed this one: `updated(key, value)` would be better IMO. I'll fix when I'll cherry-pick.

pwielgolaski: I fixed it and pushed it.

slandelle:
cherry-picked, thanks!

PR #3264, opened 2017-03-15
Prefix http names with method name in reports

hxtpoe:
This automatically adds methods type prefix to http request name. 

`http("list of sth").get() `

will be reported as : [GET] list of sth


hxtpoe:
It breaks somehow PollerActorSpec.scala 
@slandelle could you take a look?

slandelle:
@hxtpoe Thanks for contributing.

Then, I'm not super convinced:
* I don't think this feature would be helpful for lots of users. It feels very specific to the way you want your tests to work.
* that's something you can achieve with your own DSL on top of Gatling's. You don't need to have it built-in.

I'll close it for now. I'll re-open if there's some traction from the community.

Thanks anyway for your effort!

hxtpoe:
> you can achieve with your own DSL on top of Gatling's.

Yes, but this would be too easy!
I created my own HttpDsl and it fully meets my exceptations. @slandelle Thanks for suggesting another approach.

slandelle:
:)

PR #3265, opened 2017-03-16
Recorder: Redirections shoud not be generated as resources

t-takatsuru:
Redirections should not be generated as resources.
(But, It may be necessary to examine other influences.)

HTTP Referer header problem occurs in recorder, both by proxy and from HAR.

Recording in following condition:
* No follow redirects
* Automatic Referers

Redicections are generated as resources:
```scala
	val scn = scenario("RecordedSimulationNoFollowRedirects")
		.exec(http("request_0")
			.get("/top.html"))
		.pause(1)
		.exec(http("request_1")
			.get("/SampleWeb")
			.resources(http("request_2")
			.get("/SampleWeb/"))
			.check(status.is(302)))
		.pause(1)
		.exec(http("request_3")
			.get("/SampleWeb/next.html"))
```

This simulation (auto-referer) produces wrong Referer value:
```
> GET /top.html HTTP/1.1
< HTTP/1.1 200 OK
< <a href="/SampleWeb">SampleWeb</a>
> GET /SampleWeb HTTP/1.1
> Referer: http://hogehoge/top.html
< HTTP/1.1 302 Found
< Location: http://hogehoge/SampleWeb/
> GET /SampleWeb/ HTTP/1.1
> Referer: http://hogehoge/top.html
< HTTP/1.1 200 OK
< <a href="/SampleWeb/next.html">Last</a>
> GET /SampleWeb/next.html HTTP/1.1
> Referer: http://hogehoge/top.html   <<----- WHY???? (should be http://hogehoge/SampleWeb/)
< HTTP/1.1 200 OK
```

This is because redirections and resources do not change Referer session variable.

This change sets are fixes that excludes redirection from resource grouping.
Using this patch, redirections are generated as following:
```scala
	val scn = scenario("RecordedSimulationNoFollowRedirects")
		.exec(http("request_0")
			.get("/top.html"))
		.pause(1)
		.exec(http("request_1")
			.get("/SampleWeb")
			.check(status.is(302)))
		.exec(http("request_2")
			.get("/SampleWeb/"))
		.pause(1)
		.exec(http("request_3")
			.get("/SampleWeb/next.html"))
```

And Referer of ``/SampleWeb/next.html`` is produced correctly.

### DIFF
--- a/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioDefinition.scala
+++ b/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioDefinition.scala
@@ -49,7 +49,8 @@ private[recorder] object ScenarioDefinition extends StrictLogging {
 
   private def filterInferredResources(requests: Seq[TimedScenarioElement[RequestElement]]): Seq[TimedScenarioElement[RequestElement]] = {
     val groupChainedRequests = requests.groupByNeighborCondition { (a, b) =>
-      b.sendTime - a.arrivalTime < ConsecutiveResourcesMaxIntervalInMillis
+      b.sendTime - a.arrivalTime < ConsecutiveResourcesMaxIntervalInMillis &&
+      !isRedirection(a)
     }
 
     groupChainedRequests.map {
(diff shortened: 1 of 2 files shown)


slandelle:
Hi @t-takatsuru and thanks again!

Actually, we need this kind of empirical algorithm to determine resources in proxy mode. We just have no other way. So I feel like your case fixes har mode but breaks proxy mode.

Then, the recorder needs some serious refactoring.
I'd like to have both proxy and har modes produce the same AST (probably to the har one) that would then be handed over to the templates.
I'd rather keep this PR around until we do this refactoring.

t-takatsuru:
OK, you are planning refactoring, I'll wait for it :)
Just in my idea, I'd like to customize AST manipulation algorithm somehow on recording.

slandelle:
> OK, you are planning refactoring, I'll wait for it :)

Well, no ETA for that atm. So busy with things with higher priority. Feel free to give it a try :)

slandelle:
@t-takatsuru Let me close this for now. I've started working on a complete re-write of the Recorder's HTTP events AST and all redirections will be properly filtered out.

Thanks!

Issue #2602, opened 2015-03-10
Core: Provide a "streaming" CSV reader


pdalpra:
There are cases where users are loading massive CSV files as feeders for their simulations.
Currently, as CSV feeders are loaded eagerly to avoid file IO during the simulation, users may require huge heaps so that the data can fit in memory.

An alternative could be to read CSV files in small batches, which would trade memory usage for some file IO when a new batch is read.

slandelle:
linked to #3174

PR #3268, opened 2017-03-23
Provide a lazy, batched CSV reader

reikje:
see https://github.com/gatling/gatling/issues/2602

Currently csv files are loaded eagerly which is a problem when the files are very big. This `Feeder` processes portions of the csv file on demand in batches. When creating, you can specify batch size and how often you want to move to the next batch.

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/feeder/FeederBuilder.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/feeder/FeederBuilder.scala
@@ -15,6 +15,8 @@
  */
 package io.gatling.core.feeder
 
+import java.util.concurrent.locks.ReentrantLock
+
 import io.gatling.core.structure.ScenarioContext
 
 trait FeederBuilder[T] {
@@ -23,6 +25,44 @@ trait FeederBuilder[T] {
 
 case class FeederWrapper[T](feeder: Feeder[T]) extends FeederBuilder[T] {
   def build(ctx: ScenarioContext) = feeder
+}
+
+case class BatchedFeederBuilder[T](
+    records:   Iterator[Record[T]],
+    batchSize: Long,
+    strategy:  FeederStrategy      = Queue
+) extends FeederBuilder[T] {
+
+  override def build(ctx: ScenarioContext): Feeder[T] = {
+      def fetchNextBatch(): Feeder[T] = {
+        val batchRecords = (1L to batchSize).toIterator.zip(records).map {
+          case (_, record) => record
+        }.toIndexedSeq
+        strategy.feeder(batchRecords, ctx)
+      }
+
+    var currentBatch: Feeder[T] = fetchNextBatch()
+
+    var invocations = 1
+    val lock = new ReentrantLock()
+
+    new Feeder[T] {
+      def hasNext: Boolean = {
+        lock.lock()
+        try {
+          if (invocations % batchSize == 0) {
+            currentBatch = fetchNextBatch()
+          }
+          invocations += 1
+        } finally {
+          lock.unlock()
+        }
+
+        currentBatch.hasNext
+      }
+      def next: Record[T] = currentBatch.next()
+    }
+  }
 }
 
 case class RecordSeqFeederBuilder[T](
--- a/gatling-core/src/main/scala/io/gatling/core/feeder/FeederBuilder.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/feeder/FeederBuilder.scala
@@ -15,7 +15,7 @@
  */
 package io.gatling.core.feeder
 
-import java.util.concurrent.locks.ReentrantLock
+import java.util.concurrent.atomic.{ AtomicInteger, AtomicLong }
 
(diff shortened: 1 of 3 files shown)


slandelle:
Hi! Thanks for contributing!

Actually, I wonder if it really makes sense to implement lazy file based feeders and support any strategy other than queue.

Also, my concern with the implementation you suggest is the blocking that happens when fetching the next batch. Next batch(es) could be fetch in the background, and could be triggered/paused by some watermarks.

Finally, one of the problem with Feeder footprint is that most boil down to `Array[Map[String, String]]`. We could change some implementations in Gatling 3 so that we only compute the `Map`s when pulling data.
One game changer could also be Java 9 Compact Strings coming, as `String` will no longer be backed by an UTF-16 char array.

reikje:
> .. the blocking that happens when fetching the next batch

It's actually not blocking. For one single call to `hasNext` the condition `if (itemsUsed.incrementAndGet() == minUsedItemsPerBatch) {` will evaluate to `true` and this Thread will go on and fetch the next batch. While it is doing this, subsequent requests will still run on the old batch - all until the updating Thread has updated `currentBatch`.

> .. support any strategy other than queue

The strategy is only applied onto the items in the current batch (see `BatchedFeederBuilderSpec`). I agree, Queue would be the only strategy that makes sense for the whole dataset. It's currently not implemented like this.

slandelle:
> It's actually not blocking

Let me correct myself: the blocking happens when the underlying `InputStream`'s buffer runs out so it has to fetch from disk again: 

```
SeparatedValuesParser.stream(res.inputStream, CommaSeparator, quoteChar, escapeChar)
```

reikje:
I see. How is `hasNext` and `next` used currently? Is there a some synchronisation/lock that allows only 1 thread at the same time? I can't see how it would be thread-safe otherwise. In that case your suggestion to eagerly fetch the next batch would be beneficial. Let me know if I should adapt the pull request to do that.

slandelle:
See `FeedBuilder` sources: each Feeder instance is bound to a single Actor, so Iterator processing is threadsafe and sequential. And indeed, your AtomicInteger/AtomicLong are useless.

reikje:
Yes, I was thinking this code was executed in parallel. How would you implement the background loading in that case?

reikje:
@slandelle how can we move forward with this PR, it's really something we wanna use here :) any suggestion on the background loading (since it's an Actor)?

slandelle:
I guess background loading can come in a second time.

But there are several things I'm not fond of.

* `minUsedItemsPerBatch`: I fail to see why users would want to use that, except for working around some limitations of this implementation. Let's just read again once batch is depleted.
* you're leaking a file descriptor: the underlying `InputStream` is never closed
* strategies are broken. They should behave in a similar fashion as the non batched versions:
  * `circular` should open a new InputStream and start reading the file over again
  * `shuffle` should pick a random entry in the batch, but should stop when all data is depleted.
  * `random` should pick a random entry in the batch, but start reading the file over again like `circular`

reikje:
* `minUsedItemsPerBatch`: let's say we didn't have this and ".. read again once the batch is depleted". What is your definition of "depleted"? That I have exhausted my batch records via the `FeederStrategy`? `random` and `circular` never exhaust (`hasNext` will always be true). `minUsedItemsPerBatch` simply defines at which point the next batch should be loaded regardless of the selected `FeederStrategy`. So for instance, a user may chose to get 1000 items from the current batch of size 100 using the `Circular` strategy and then move on to the next batch.

* leak: will fix

* broken strategies: when the batch is initialised in `def nextBatch()` it delegates to the existing strategies, so it should work very much the same way (per batch strategy). Sorry I [don't see](https://github.com/gatling/gatling/blob/master/gatling-core/src/main/scala/io/gatling/core/feeder/FeederStrategy.scala) where `circular` or `random` reads the underlaying file again. It looks like the only difference are the access patterns to the given `records: IndexedSeq[Record[T]]`.

slandelle:
You're focused on the implementation and reusing the current implementation strategies. There's a good chance we'll have to change them.

As a user, I don't care if the same piece of code is involved for batch and non-batched or not. What I care about is behavior. Having a circular behavior is completely unrelated to batch/lazy vs full/eager loading.

reikje:
Yes `circular` and `queue` could easily work the same way between batched and non-batched. `shuffle` and `random` are not as straight-forward, or how would you implement these when you have 1M entries and 100 batches of 10.000 items?

slandelle:
> how would you implement these when you have 1M entries and 100 batches of 10.000 items

I think it would be fine if both would shuffle the batch when it's being loaded.
On depletion,`random` would jump back to the beginning of the file while `shuffle` would exit, just like `queue`.

slandelle:
@reikje Sorry for the very long delay. I just pushed my own take on batched feeders, inspired by your version: ccac0dfc18a4f48281670cc1c7fc4f3a0076cdfd

I didn't try to handle possible hiccups when reading next batch, will have to measure.
Feedback welcome.

Thanks again!

PR #3279, opened 2017-04-12
Fix SplitInjection to alse work with only one split, close #3278

### LINKED ISSUE #3278: Split users when possibleUsers == stepsUsers
When using SplitInjection with only enough users to make one cycle, no users are injected.

This is pretty minor but can be annoying when we want to let a user choose how many cycle (split) he wants to run.

I will soon propose a patch.

frol2103:


slandelle:
cherry picked

PR #3283, opened 2017-04-19
Loop as long as with duration limit

wiacekm:
I think it would be good to add another type of loop statement where we iterate over the loop as long as condition is satisfied but no longer than specified amount of time.
It would allow to avoid some tricky solution and hacks when we have to wait for some data in response. And we cannot predict when it will be populated (ex due to caching).

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/Runner.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/Runner.scala
@@ -84,7 +84,7 @@ private[gatling] class Runner(system: ActorSystem, configuration: GatlingConfigu
     simulation.executeBefore()
     logger.trace("Before hooks executed")
 
-    val runMessage = RunMessage(simulationParams.name, selection.userDefinedSimulationId, selection.defaultSimulationId, nowMillis, selection.description)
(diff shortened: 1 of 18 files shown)


gatling-core/src/main/scala/io/gatling/core/action/builder/LoopBuilder.scala.orig
```diff
@@ -0,0 +1,59 @@
...
+case object DoWhileType extends LoopType("doWhile", false, true)
+=======
+sealed abstract class LoopType(val name: String, val timeBased: Boolean)
+case object RepeatLoopType extends LoopType("repeat", false)
+case object ForeachLoopType extends LoopType("foreach", false)
+case object DuringLoopType extends LoopType("during", true)
+case object ForeverLoopType extends LoopType("forever", false)
+case object AsLongAsLoopType extends LoopType("asLongAs", false)
+case object AsLongAsDuringLoopType extends LoopType("asLongAsDuring", true)
+>>>>>>> Stashed changes
```
slandelle: Broken merge conflict.

wiacekm:
hopefully everything is ok now

wiacekm:
let me know if I should open new pull request or reopen this one

slandelle:
cherry-picked, thanks!

Issue #3296, opened 2017-05-18
Core: Json.Stringify does not escape double quotes


jozemlakar:
io.gatling.core.json.Json.stringify method stringifies objects as JSON strings. This generally works fine, but not when a string value has double quotes. This would seem safe generally, but not when a whole JSON is stringified. Imagine this case:

```
val mapOriginalValue = Map[String,Any]("key1" -> "Regular foo", "key2" -> "Quoted \"foo\"");
val stringifiedMap = stringify(mapOriginalValue);
println("Stringified JSON: " + stringifiedMap);
```

expected output would be 
`Stringified JSON: {"key1":"Regular foo","key2":"Quoted \"foo\""}`
instead the result is
`Stringified JSON: {"key1":"Regular foo","key2":"Quoted "foo""}`

This is not valid JSON. 
I have tried to find where the problem is. In file [Json.scala](https://github.com/gatling/gatling/blob/master/gatling-core/src/main/scala/io/gatling/core/json/Json.scala) line 48, escaping is implemented for line feeds but not double quotes. 

Adding escaping for double quotes is easy and I can prepare a PR, but this might break existing code, depending on this "feature". You can recommend the proper solution and I can implement it. 

```
val originalValue = "Quoted \"foo\"";
val expectedValue = "Quoted \\\"foo\\\"";
val stringified = stringify(originalValue);
if (stringified == expectedValue) {
	println("Stringified works properly")
} else {
	println("Stringify is not working properly")					
}
```

results in `Stringify is not working properly`

slandelle:
Thanks for reporting!
Are you interested in contributing a fix?

jozemlakar:
yes. I am testing a fix right now, but I am not sure if maybe someone somewhere is using this code in a wrong way and the fix will break their code

jozemlakar:
Added PR https://github.com/gatling/gatling/pull/3297 to fix this

jozemlakar:
Replaced with PR #3298

PR #3297, opened 2017-05-19
Adding escaping of double quotes and tests #3296

jozemlakar:
I have added the escaping and two tests (simple string value and a map)

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
@@ -45,7 +45,7 @@ object Json {
   }
 
   private def writeString(s: String, rootLevel: Boolean) = {
-    val escapedLineFeeds = s.replace("\n", "\\n")
+    val escapedLineFeeds = s.replace("\n", "\\n").replace("\"", "\\\"")
     if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""
   }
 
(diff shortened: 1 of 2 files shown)


gatling-core/src/main/scala/io/gatling/core/json/Json.scala
```diff
@@ -45,7 +45,7 @@ object Json {
   }
 
   private def writeString(s: String, rootLevel: Boolean) = {
-    val escapedLineFeeds = s.replace("\n", "\\n")
+    val escapedLineFeeds = s.replace("\n", "\\n").replace("\"", "\\\"")
```
slandelle: I don't think that's sufficient. What if content is already escaped?

jozemlakar:
I agree. Should we implement a full escaping system or find an existing one? This fix is for this specific problem and it works. The question is... is it enough? What if \n was already escaped?

slandelle:
> Should we implement a full escaping system or find an existing one?
What if \n was already escaped?

The main problem is we can't know if content is already escaped or not. And I agree there's the same issue if `\n` was already escaped. I'm not sure we'll find a real escaping system that would be smart enough to detect pre-escaped content. We have the same level of complexity for url escaping in gatling-http/AsyncHttpClient.

A real escaping engine that would comply with the full spec would have to support unicode and other weird things. It's definitively beyond the scope here.

I'd be in favor of fix the possible double escaping of double quotes and line feeds and period.

jozemlakar:
I hate to reinvent the wheel so I went and checked what is already available in Gatling. I am familiar with Jackson so I went and used that and it works. Is this the way to go? Should we allow the use of Boon like in parser?

slandelle:
Go with Jackson, Boon is a bit stalled atm.

PR #3298, opened 2017-05-19
Replaced manual escaping with Jackson

jozemlakar:
1. I have removed the manual escaping of new line
2. Escaping is now done by Jackson (only used for String)
3. ObjectMapper is now created when first used. According to this [article](http://stackoverflow.com/questions/3907929/should-i-declare-jacksons-objectmapper-as-a-static-field) it is thread-safe
4. Jackson adds double quotes to start and end of the stringified String object. These are removed using regex to fit the existing code.

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
@@ -21,6 +21,8 @@ import scala.collection.JavaConverters._
 
 import com.dongxiguo.fastring.Fastring.Implicits._
 
+import com.fasterxml.jackson.databind.ObjectMapper
+
 object Json {
 
   def stringify(value: Any, isRootObject: Boolean = true): String =
@@ -45,8 +47,9 @@ object Json {
   }
 
   private def writeString(s: String, rootLevel: Boolean) = {
-    val escapedLineFeeds = s.replace("\n", "\\n").replace("\"", "\\\"")
-    if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""
+    val objectMapper = new ObjectMapper
+    val escapedLineFeeds = objectMapper.writeValueAsString(s).replaceAll("^\"|\"$", "")
+    if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""   
   }
 
   private def writeValue(value: Any) = fast"${value.toString}"
--- a/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
@@ -50,7 +50,7 @@ object Json {
 
   private def writeString(s: String, rootLevel: Boolean) = {
     val escapedLineFeeds = objectMapper.writeValueAsString(s).replaceAll("^\"|\"$", "")
-    if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""   
+    if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""
   }
 
   private def writeValue(value: Any) = fast"${value.toString}"
--- a/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/json/Json.scala
@@ -45,7 +45,7 @@ object Json {
   }
 
   private def writeString(s: String, rootLevel: Boolean) = {
-    val escapedLineFeeds = s.replace("\n", "\\n")
+    val escapedLineFeeds = s.replace("\n", "\\n").replace("\"", "\\\"")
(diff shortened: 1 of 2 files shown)


jozemlakar:
I have no idea why the build failed. Feeder timeout...?

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
```diff
@@ -45,8 +49,8 @@ object Json {
   }
 
   private def writeString(s: String, rootLevel: Boolean) = {
-    val escapedLineFeeds = s.replace("\n", "\\n")
-    if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""
+    val escapedLineFeeds = objectMapper.writeValueAsString(s).replaceAll("^\"|\"$", "")
```
slandelle: Why do you need to replace some chars?

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
```diff
@@ -45,8 +49,8 @@ object Json {
...
+    if (rootLevel) fast"$escapedLineFeeds" else fast""""$escapedLineFeeds""""
```
slandelle: trailing spaces

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
jozemlakar: Jackson (correctly) adds double quotes to string as it is stringified.
`This is my string` will result in `"This is my string"` 

The purpose is of course that it aims to achieve the final goal being:
`key: "This is my string"` 

However in gatling double quotes are added later (ie writeMap) irrespective of the type. And so we would end up with `key: ""This is my string""` which is wrong. The reqex removes the double quoted at both ends.

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
jozemlakar: Removed.

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
```diff
@@ -21,10 +21,14 @@ import scala.collection.JavaConverters._
...
 import com.dongxiguo.fastring.Fastring.Implicits._
 
+import com.fasterxml.jackson.databind.ObjectMapper
+
 object Json {
 
   def stringify(value: Any, isRootObject: Boolean = true): String =
     fastringify(value, isRootObject).toString()
+  
+  private val objectMapper = new ObjectMapper
```
slandelle: It saddens me to have a mutable stateful singleton :(
We've worked hard in Gatling 2 to remove those that were making testing very difficult and brittle.

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
```diff
@@ -45,7 +49,7 @@ object Json {
   }
 
   private def writeString(s: String, rootLevel: Boolean) = {
-    val escapedLineFeeds = s.replace("\n", "\\n")
+    val escapedLineFeeds = objectMapper.writeValueAsString(s).replaceAll("^\"|\"$", "")
```
slandelle: It's really sad we have to reprocess the Jackson output :(

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
jozemlakar: I agree, The alternative would have been to create the mapper for the instance. I will try and fix this.

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
jozemlakar: I absolutely agree. And I believe that Jackson got it right. We should fix the way gatling build JSON probably. I tried to keep the change contained. Please decide.

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
slandelle: Don't do that, recreating the ObjectMapper on every call is a perf killer :(

gatling-core/src/main/scala/io/gatling/core/json/Json.scala
slandelle: > We should fix the way gatling build JSON probably.

We can't do that. It's intended to support re-injecting a captured JSON tree in a node position in a Gatling EL template.

jozemlakar: I agree with all of above. The only way to get out of this mess is to copy the escaping mechanism from Jackson to gatling. Or to find one in Scala.

slandelle:
@jozemlakar I'm very sorry with the very long delay :(

I'm going to cherry-picked this and build on top of it

Thanks!

PR #3308, opened 2017-05-26
Add the code and fix the anchor of sse().reconciliate

mwpeng:
Add the code and fix the anchor of sse().reconciliate


slandelle:
Cherry-picked. Thanks a lot!

PR #3309, opened 2017-05-26
Fix feeder shuffle description

joemeszaros:
# Problem 

Feeder shuffle description had a typo:

`shuffle entries, then behave live queue`

# Solution

Change `live` -> `like`


slandelle:
Thanks! I had already fixed on our private branch, but I'll cherry-pick yours instead so you're credited with the commit :)

joemeszaros:
:-) thanks

slandelle:
cherry-picked

PR #3339, opened 2017-08-03
Unify the bracket usage

Fokko:
Hi all,

I know I'm a bit of a formatting purist, but I would like to strip the brackets to get to a single unified way of formatting. The brackets at the `start` function are redundant.

Cheers, Fokko


slandelle:
Oh my. We do hunt those down but missed this one, thanks! :)

cherry-picked

PR #3346, opened 2017-08-25
normal builder for GatlingPropertiesBuilder

mskonovalov:
After fromArgs became private the only way to run simulation programmatically is to use GatlingPropertiesBuilder. So it'd be better to make it real builder to provide fluent API

### DIFF
--- a/gatling-charts/src/test/scala/io/gatling/charts/result/reader/LogFileReaderSpec.scala
+++ b/gatling-charts/src/test/scala/io/gatling/charts/result/reader/LogFileReaderSpec.scala
@@ -16,7 +16,6 @@
 package io.gatling.charts.result.reader
 
 import scala.collection.mutable
-
 import io.gatling.BaseSpec
 import io.gatling.charts.stats.LogFileReader
 import io.gatling.core.ConfigKeys._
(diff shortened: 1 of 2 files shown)


slandelle:
Why not. But then, could you please fix formatting? I'd say you've just formatted with IntelliJ, please compile with sbt, it will trigger formatting.

> After fromArgs became private the only way to run simulation programmatically is to use GatlingPropertiesBuilder.

What are you doing exactly? We intend Gatling as a software, not a library, so the public API is the DSL, not this class, which is subject to change.

mskonovalov:
You know, when you write your scenarios with IDE it is much easier to run as a code with debug instead of running as a sbt or sh script
So INHO it is very important to provide such API for programmatic launch

slandelle:
For maven, as it doesn't have an interactive mode, I agree.
That's why we provide helpers generated by the maven-archetype and also placed in the [maven demo](https://github.com/gatling/gatling-maven-plugin-demo/blob/master/src/test/scala/Engine.scala).

IMO, sbt users don't really need that. sbt interactive mode with autocompletion and testOnly is way more convenient.

mskonovalov:
frormatting fixed

slandelle:
Have you pushed? formatting is still broken in LogFileReaderSpec.

mskonovalov:
Yes,
strange but sbt compile doesn't change this file

slandelle:
Because "compile" sbt task only compiles "main" sources.
As it's a test class, you have to run "test:compile".

mskonovalov:
Oh, sure, sorry  missed that

slandelle:
cherry-picked

thanks!

Issue #3331, opened 2017-07-05
Verbose INFO logging when warm up fails


EugeneAbramchuk:
Gatling version 2.2.4

If Gatling fails to warm up from the first time, it throws an exception, which is unnecessary (it could be handled).

It would be cleaner if Gatling reported:

11:34:14.605 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling.io after 1000 ms
11:34:14.xxx [INFO ] i.g.h.a.HttpEngine - Warm up done

```

11:34:14.605 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling.io
java.util.concurrent.TimeoutException: Request timeout to not-connected after 1000 ms
	at org.asynchttpclient.netty.timeout.TimeoutTimerTask.expire(TimeoutTimerTask.java:43)
	at org.asynchttpclient.netty.timeout.RequestTimeoutTimerTask.run(RequestTimeoutTimerTask.java:48)
	at io.netty.util.HashedWheelTimer$HashedWheelTimeout.expire(HashedWheelTimer.java:625)
	at io.netty.util.HashedWheelTimer$HashedWheelBucket.expireTimeouts(HashedWheelTimer.java:700)
	at io.netty.util.HashedWheelTimer$Worker.run(HashedWheelTimer.java:428)
	at java.lang.Thread.run(Thread.java:745)
Wrapped by: java.util.concurrent.ExecutionException: java.util.concurrent.TimeoutException: Request timeout to not-connected after 1000 ms
	at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1895)
	at org.asynchttpclient.netty.NettyResponseFuture.get(NettyResponseFuture.java:168)
	at io.gatling.http.ahc.HttpEngine.warmpUp(HttpEngine.scala:96)
	at io.gatling.http.protocol.HttpProtocol$$anon$1.io$gatling$http$protocol$HttpProtocol$$anon$1$$$anonfun$1(HttpProtocol.scala:62)
	at io.gatling.core.protocol.ProtocolComponentsRegistry.comps$1(Protocol.scala:67)
	at io.gatling.core.protocol.ProtocolComponentsRegistry.io$gatling$core$protocol$ProtocolComponentsRegistry$$$anonfun$4(Protocol.scala:69)
[...]

PR #3355, opened 2017-09-24
Added the clause to hide HTTP engine warm up exceptions from log level higher than debug, close gatling/gatling#3331

EugeneAbramchuk:
Info log level now shows no exception and provides the timeout in the message for clarity (previously not stated):
16:53:20.762 [INFO ] i.g.h.a.HttpEngine - Start warm up
16:53:20.850 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling.io after 1000 ms
16:53:20.852 [INFO ] i.g.h.a.HttpEngine - Warm up done

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
@@ -73,6 +73,8 @@ class HttpEngine(
 
       import httpComponents._
 
+      val debugEnabled = logger.underlying.isDebugEnabled
+
       if (httpProtocol.enginePart.perUserNameResolution) {
         // eager load
         val _ = defaultDnsNameResolver
@@ -80,20 +82,25 @@ class HttpEngine(
 
       httpProtocol.warmUpUrl match {
         case Some(url) =>
+          val connectionTimeout: Int = 1000
           val requestBuilder = new RequestBuilder().setUrl(url)
             .setHeader(Accept, "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
             .setHeader(AcceptLanguage, "en-US,en;q=0.5")
             .setHeader(AcceptEncoding, "gzip")
             .setHeader(Connection, KeepAlive)
             .setHeader(UserAgent, "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:16.0) Gecko/20100101 Firefox/16.0")
-            .setRequestTimeout(1000)
+            .setRequestTimeout(connectionTimeout)
 
           httpProtocol.proxyPart.proxy.foreach(requestBuilder.setProxyServer)
 
           try {
             ahcFactory.defaultAhc.executeRequest(requestBuilder.build).get
           } catch {
-            case NonFatal(e) => logger.info(s"Couldn't execute warm up request $url", e)
+            case NonFatal(e) => if (debugEnabled) {
+              logger.debug(s"Couldn't execute warm up request $url", e)
+            } else {
+              logger.info(s"Couldn't execute warm up request $url after {} ms", connectionTimeout)
+            }
           }
 
         case _ =>
--- a/gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
@@ -27,6 +27,7 @@ import io.gatling.http.protocol.{ HttpComponents, HttpProtocol }
 import io.gatling.http.request.builder.Http
 import io.gatling.http.resolver.{ CacheOverrideNameResolver, ExtendedDnsNameResolver }
 import io.gatling.http.util.HttpTypeCaster
+import io.gatling.commons.util.Throwables._
 
 import akka.actor.ActorSystem
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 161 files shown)


gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
```diff
@@ -73,27 +73,33 @@ class HttpEngine(
 
       import httpComponents._
 
+      val debugEnabled = logger.underlying.isDebugEnabled
```
slandelle: this code will be only executed once, hence no need to have a val for this, just directly test `logger.underlying.isDebugEnabled` below

gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
```diff
@@ -73,27 +73,33 @@ class HttpEngine(
...
+
       if (httpProtocol.enginePart.perUserNameResolution) {
         // eager load
         val _ = defaultDnsNameResolver
       }
 
       httpProtocol.warmUpUrl match {
         case Some(url) =>
+          val connectionTimeout: Int = 1000
```
slandelle: no need for a val here, it won't be reused. please just leave it as it was

gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
```diff
@@ -73,27 +73,33 @@ class HttpEngine(
...
           httpProtocol.proxyPart.proxy.foreach(requestBuilder.setProxyServer)
 
           try {
             ahcFactory.defaultAhc.executeRequest(requestBuilder.build).get
           } catch {
-            case NonFatal(e) => logger.info(s"Couldn't execute warm up request $url", e)
+            case NonFatal(e) => if (debugEnabled)
+              logger.debug(s"Couldn't execute warm up request $url", e)
+            else
+              logger.info(s"Couldn't execute warm up request $url after {} ms", connectionTimeout)
```
slandelle: directly inject `connectionTimeout` in a placeholder in the interpolated String, just like `$url`.

gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
EugeneAbramchuk: It was hardcoded before, I introduced a val to output it with the info level message. Do you want me to return the hardcoded `.setRequestTimeout(1000)`? (You suggested to inject `connectionTimeout ` in the next comment)

slandelle: Well, the way you've changed it is still hardcoded (it can't be configured, and I fail to see people tuning a configuration option for that, most users aren't even aware of it and it's a good thing), but just at a different place. So yes, please restore it the way it was.

gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
EugeneAbramchuk: done, not sure why I introduced it in the first place

gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
slandelle: Oh, right, you use it in the log message.

gatling-http/src/main/scala/io/gatling/http/ahc/HttpEngine.scala
slandelle: Please align log messages. The only difference should be the stacktrace in DEBUG while INFO should only have the Exception message (use Throwables.detailedMessage).

EugeneAbramchuk: Do we really need `detailedMessage` in the info log level output? Throwable.getMessage is already detailed enough in my view:

> 18:10:14.093 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling111.io: java.net.UnknownHostException: gatling111.io: nodename nor servname provided, or not known

slandelle: Sorry, I don't get you. `Throwable.getMessage` doesn't report the class name, and can return null if the Throwable wasn't created with a message.

EugeneAbramchuk: Sorry for the confusion, the method names misled me into thinking they are more detailed than they really are. Throwable.getMessage has more or less the same output as Throwables.detailedMessage.

It's done now:

```
11:53:13.747 [INFO ] i.g.h.a.HttpEngine - Start warm up
11:53:13.942 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling111.io: j.u.c.ExecutionException: java.net.UnknownHostException: gatling111.io: nodename nor servname provided, or not known
11:53:13.944 [INFO ] i.g.h.a.HttpEngine - Warm up done
```

slandelle:
LGTM :)

EugeneAbramchuk:
Yay! Looks like there is the last formatting conflict (spaces replaced with tabs or the likes), not sure where it originated from - I don't think I committed the file.

slandelle:
chery-picked as of b175e08adc5b9e4f702b6947d410955c9d04f1a1

Thanks!

EugeneAbramchuk:
if `e.detailedMessage` is swapped for `e.getCause.getMessage`, it further simplifies the output from

15:56:51.941 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling111.io: j.u.c.ExecutionException: java.net.UnknownHostException: gatling111.io: nodename nor servname provided, or not known

to 

15:52:00.155 [INFO ] i.g.h.a.HttpEngine - Couldn't execute warm up request http://gatling111.io: gatling111.io: nodename nor servname provided, or not known


Sorry for nagging, I wondered if this is what you meant the change to really be.

slandelle:
`detailedMessage` is not available in 2.3, hence using `rootMessage` instead. The class name information has some value, so I'd rather have it in master.

PR #3377, opened 2017-11-11
warn against initializing loop counter

mz0:
and explain consequences


slandelle:
Thanks. Looks like a good idea. I'll cherru-pick and then rephrase a bit.

slandelle:
cherry-picked

PR #3382, opened 2017-11-15
HTTP: Add support for secure cookies in HTTP helpers, close #3381.

### LINKED ISSUE #3381: Handle Secure Cookies
Adding a secure cookie is not possible currently. I want to add this possibility. (Pull request on its way). Next to that I want to configure the session to also be able to handle a secure cookie on a http domain.

Make sure these checkboxes are checked before submitting your issue. Thanks you!

* [x] Don't post questions here. This bugtracker is for issues and feature requests only. Use the [community mailing list](https://groups.google.com/forum/#!forum/gatling) instead.
* Regarding issues:
  * [x] Check that you can reproduce your issue with latest Gatling version
  * [x] Search this bugtracker for any similar issue that might have already been reported (and possibly fixed and available as a SNAPSHOT version)
  * [x] Provide a way for Gatling developers to reproduce your problem, e.g. step-by-step instructions against a public facing website, or a sample application that demonstrates your problem. See added test scenarios
* Regarding feature requests:
  * [ ] Explain your use case, not the implementation you have in mind.

krisgeus:


slandelle:
Hi,

Thanks!
Indeed, the ability to specify the `Secured` attribute on `addCookie` and `getCookie` is missing, thanks for contributing.

Then, I don't get what this "force secure" is about. You seem to try to control the default `Secured` value with a Session attribute. I guess you plan on setting it yourself manually. This looks wrong to me:

* Why would you have to resolve such default per virtual user?
* The proper way to define default values is either in the protocol or in the config
* You're not supposed to hack onto Gatling private attributes, they are for internal use only. There's a very good chance we're going to make them private so you won't be able to use them in your simulation.
* If you're just trying to avoid repeating `withSecured(true)` on all your `add/getCookie` actions (do you have that many), why don't you just create a helper method on top of Gatling DSL?

Except if you manage to convince me that this makes sense, I would rather have it removed.

Finally, could you please update the documentation (`http_helper.rst`)?

Thanks again!

krisgeus:
Hi. In our specific case we are running a integration test (with gatling) against our acceptance environment. The api endpoint is http but is behind nginx which provides the https part. The cookie in that environment will be secure. So will the cookie we need to set from within the simulation.

So far so good. But the default getstoredcookies method will determine the cookie to retrieve base on the scheme of the uri. So the secured cookie is not send with the actual request if the base url is http.
The session variable to force retrieving of the secured cookie is therefore needed.
I used the private prefix but of course that is not needed, so I will at least remove that. 

I hope I convinced you to keep the forceSecured mechanism.
I will add some documentation around this in the proposed documentation page.

slandelle:
> The api endpoint is http but is behind nginx which provides the https part.

> So the secured cookie is not send with the actual request if the base url is http.

Sorry, I don't get it. Are you saying your API always generates secured cookies and you're trying to bypass the Nginx that's supposed to expose the API over https?
This looks super hacky to me. Why not have an option on your API that would disable generating secured cookies?

krisgeus:
Well you understood it very well ;-). I agree it's a bit hacky yes. I will remove the hacky bit about forcing the secure cookie with http url's and run my integration tests through nginx.

slandelle:
Yes, that would be better :)

BTW, IMHO, your API shouldn't generate secured cookies if it runs over plain HTTP.
As Nginx is responsible to upgrading to https, it should be the one in charge of with rewriting and securing the cookies:
* https://unix.stackexchange.com/questions/306463/nginx-add-secure-flag-to-cookies-from-proxied-server
* https://github.com/AirisX/nginx_cookie_flag_module

krisgeus:
Thx for the tip. I will have a look into them. 
Pull request should be ready now ;-)

gatling-http/src/main/scala/io/gatling/http/cookie/CookieSupport.scala
```diff
@@ -35,11 +35,13 @@ object CookieSupport {
 
   def getStoredCookies(session: Session, url: String): List[Cookie] = getStoredCookies(session, Uri.create(url))
 
-  def getStoredCookies(session: Session, uri: Uri): List[Cookie] =
+  def getStoredCookies(session: Session, uri: Uri): List[Cookie] = {
+
```
slandelle: please remove empty line and useless {}

krisgeus:
Done

slandelle:
cherry-picked as 408cf8d65b6739d79c5d7858f39d6bb7667ebae2

Thanks!

Issue #3386, opened 2017-11-21
HTTP: Redirected request is accounted twice in cumulatedResponseTime


kudoueiji:
I checked the operation with the latest source(master branche at 2017/11/20)

I set FollowRedirect is enabled and
define a scenario consisting only of groups containing one request.
(group name is　'group1', request name is 'request1' ) 
After receiving the response of request 1, request 'request1 Redirect 1' is generate.

I execute this simulation and expect the following results.
If 'request1' response time is A and  'request1 Redirect 1' response time is B,
'group1' cumulatedResponseTime is A+B.
But cumulatedResponseTime is A+A+B.

The whole file I used is below.
It is necessary to change the proxy settings to use.
[CD.scala.txt](https://github.com/gatling/gatling/files/1490221/CD.scala.txt)

slandelle:
Thanks for reporting!

PR #3387, opened 2017-11-21
When followRedirect is enable, groupUpdate do not relay.

kudoueiji:
#3386


slandelle:
I'm sorry, I only noticed your PR after committing a fix for your issue.
Actually, we shouldn't propagate updates when we're not fetching a resource.

Thanks anyway!

PR #3389, opened 2017-11-22
On the Statistics chart, sort group names and request names by name.

kudoueiji:
This is a suggestion.
In the Statistics chart, it is easy to see that group names and request names are sorted by name.

For example:
example


slandelle:
Hi,
Actually, default/initial sort order is by first occurrence time asc, which really makes sense as it reflects the execution order in your workflow.
I'm inclined to leave it as is.

kudoueiji:
Since the display order of requests included in '.resources'  is changed every time,
I thought it was a random order.

I think that the default / initial sort order is good in ascending order of the first occurrence time.
I will cancel this proposal.

slandelle:
Thanks for your feedback.

PR #3401, opened 2017-12-13
Tracker Destination can by specified independent from replyTo Destination

LukasHavemann:
Hello, 

we have a complex JMS setup in which we need to specify the tracker destination independent from the replyTo destination.

Best regards
Lukas

### DIFF
--- a/gatling-jms/src/main/scala/io/gatling/jms/action/RequestReply.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/action/RequestReply.scala
@@ -32,14 +32,16 @@ import io.gatling.jms.request._
  * This handles the core "send"ing of messages. Gatling calls the execute method to trigger a send.
  * This implementation then forwards it on to a tracking actor.
  */
-class RequestReply(attributes: JmsAttributes, replyDestination: JmsDestination, protocol: JmsProtocol, jmsConnectionPool: JmsConnectionPool, val statsEngine: StatsEngine, val next: Action)
+class RequestReply(attributes: JmsAttributes, replyDestination: JmsDestination, trackerDestination : Option[JmsDestination], protocol: JmsProtocol, jmsConnectionPool: JmsConnectionPool, val statsEngine: StatsEngine, val next: Action)
   extends JmsAction(attributes, protocol, jmsConnectionPool) {
 
   override val name = genName("jmsRequestReply")
 
   private val jmsReplyDestination = jmsConnection.destination(replyDestination)
+  private val jmsTrackerDestination = trackerDestination.map(dest => jmsConnection.destination(dest))
+
+  private val tracker = jmsConnection.tracker(jmsTrackerDestination.getOrElse(jmsReplyDestination), attributes.selector, messageMatcher)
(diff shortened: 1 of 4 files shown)


slandelle:
Sorry, I don't get it. Could you please explain your use case, how the current features don't let you do that and how you're trying to solve it?

LukasHavemann:
Our use case looks like this:

System A is sending a inital request to system B.
System B answers the request and is waiting for a confirmation from system A. 
System A sends the confirmation  to system B. 
System B sends a notfication to system C. 

We now want to loadtest system B and want to measure the runtime between the inital request and the notification to system C.

So we use Gatling to initiate the request and to receive the notification to system C.  We have a Stub   that receives the response from System B and sends the confirmation. 

Thats why we need to specify the replyTo independently from the queue where Gatling is listing for the response. If our use case were a simple request/response scenario we wouldn't use a async protocol like JMS.

LukasHavemann:
For clarification: our very simplified testsetup looks like this:

lasttest

LukasHavemann:
I am new to scala and sbt. How can I build a new release of gatling with sbt locally?

LukasHavemann:
@slandelle any updates on this pr?

slandelle:
Can’t you have Gatling deal with both initial request and confirmation and measure overall time with a group ?

LukasHavemann:
I don't want to hesitate you with our concrete setup and all the details. To summarize: with a lot of effort, it may be possible to replicate the behaviour of teststub T in gatling, but that's not an option for us.

Decoupling, asynchronicity and routing are big advantages of async messaging systems. I think a load testing tool like gatling, should support these concepts and shouldn't limit the user to simple request/response scenarios. I can understand that you want to keep the API small, but this small change would make gatling much more useful in a complex MQ-setups.

slandelle:
Again, I don't get why you can't do:

```scala
group("total") {
exec(jms("initialRequest").requestReply.replyDestination(queue("queueB"))
.exec(jms("confirmation").requestReply.replyDestination(queue("queueC"))
}
```

The tracker is an implementation detail and it doesn't make sense to turn it into an API.

LukasHavemann:
The generation of the confirmation message is relatively CPU expensive. If we would use a group, we wouldn't be able to scale horizontally, and we need to scale the machine running gatling vertically.

It is only an internal implementation detail, if you limit the user to simple request/response scenarios. If you embrace all the use cases of an async messaging system, it isn't. Other jms load testing tools allow the user to specify the listener (tracker) destination independently from the replyTo destination. But these tools are not as fast and good as gatling.

slandelle:
Could you please address comments and add documentation (https://github.com/gatling/gatling/blob/master/src/sphinx/jms.rst)?

slandelle:
Oups, forgot to request changes so they were still pending.

opensource21:
I fixed the conflicts and use your comments to improve the code. Furthermore I added the documentation. See #3428 . Is think we could close this PR.

slandelle:
@opensource21 Thanks! Having a look at your PR right away!

PR #3405, opened 2017-12-23
Add method 'runForResults' to run Gatling and get run result

mperever:
Hello,

The main idea is provide functionality to get gatling run result for following processing. Gatling saves an execution results and simulation log in autogenerated folder. There is no way to get autogeneraed value, but thvaule is stored in 'RunResult'.
I suggest adding a method that returns the result of gatling run.

Thanks
Maksym

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/Gatling.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/Gatling.scala
@@ -33,16 +33,18 @@ import org.slf4j.LoggerFactory
  */
 object Gatling extends StrictLogging {
 
+  def runForResults(overrides: ConfigOverrides): GatlingRunResult = start(overrides, None)
+
   // used by bundle
   def main(args: Array[String]): Unit = sys.exit(fromArgs(args, None))
 
   // used by maven archetype
-  def fromMap(overrides: ConfigOverrides): Int = start(overrides, None)
+  def fromMap(overrides: ConfigOverrides): Int = start(overrides, None).statusCode.code
 
   // used by sbt-test-framework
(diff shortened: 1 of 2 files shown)


slandelle:
What' the purpose of this change?

mperever:
The purpose is to get Gatling run result object after execution, instead of execution status code (int).
https://github.com/gatling/gatling/blob/master/gatling-app/src/main/scala/io/gatling/app/RunResult.scala
https://github.com/gatling/gatling/blob/master/gatling-app/src/main/scala/io/gatling/app/cli/StatusCode.scala

RunResult allows to implement other features for Gatling.
In my project, I need use LogFileReader to get some stats, but LogFileReader requires RunResult.
https://github.com/gatling/gatling/blob/master/gatling-charts/src/main/scala/io/gatling/charts/stats/LogFileReader.scala

Furthermore I found that a lot of people is asking how to get auto generated folder name with log file.
simulation-xxxxxx\simulation.log

slandelle:
So basically, I have the feeling the change you suggest helps you with some custom launcher/integration that you have but doesn't benefit Gatling, ie the official components that we provide.

mperever:
Yes, but in general it's good to have an opportunity to get RunResult to integrate Gatling with other systems.

slandelle:
Sorry, but please understand that we manage a large ecosystem with lots of components, from open-source ones to commercial products. We have nothing to gain from opening this, and on the contrary it would bring more complexity when/if we want to change things.
Basically, what you suggest is not a feature, and I'll have to turn it down.
I hope you'll understand.

PR #3421, opened 2018-02-14
When the body size is 0 and the Transfer-Encoding header is not trans…

kudoueiji:
When the body size is 0 and the Transfer-Encoding header is not transmitted,
The POST, PUT method should send the Content-Length header (value 0).

If the POST method does not have a Content-Length header,
IIS(Microsoft Internet Information Services ) will return status code 411.

Currently, Gatling does not automatically give the Content-Length header 
even if you generate a definition of HTTP POST method without body by GatlingRecorder and execute Simulation.
I confirmed the operation at the latest source (February 15, 2018).

The following is extracted from RFC 7230

A user agent SHOULD send a Content-Length in a request message when
no Transfer-Encoding is sent and the request method defines a meaning
for an enclosed payload body.  For example, a Content-Length header
field is normally sent in a POST request even when the value is 0
(indicating an empty payload body).  A user agent SHOULD NOT send a
Content-Length header field when the request message does not contain
a payload body and the method semantics do not anticipate such a
body.

### DIFF
--- a/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
+++ b/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
@@ -17,6 +17,7 @@
 package io.gatling.recorder.scenario.template
 
 import io.gatling.commons.util.StringHelper.EmptyFastring
+import io.gatling.http.HeaderNames
 import io.gatling.http.util.HttpHelper.OkCodes
 import io.gatling.recorder.config.RecorderConfiguration
 import io.gatling.recorder.scenario.{ RequestBodyBytes, RequestBodyParams }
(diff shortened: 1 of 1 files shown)


slandelle:
Hi,

Thanks for reporting.

First, have you opened an issue against IIS too? As the RFC says SHOULD and not MUST, a server should be able to support such POST request without a Content-Length header.

Then, once again, old RFCs are poorly designed. "the request method defines a meaning
for an enclosed payload body" and "the method semantics do not anticipate such a
body" are things that are never specified anywhere. For example, GET requests can have a message body (and ElasticSearch makes use of such requests).

Finally, sorry, but your fix is wrong. Gatling HTTP support, and or AsyncHttpClient should be the ones adding such 0 ContentLength. This is not the responsibility of the recorder. Would you mind providing another fix, please?

slandelle:
Fixed in AHC as of https://github.com/AsyncHttpClient/async-http-client/issues/1521

slandelle:
Thanks for reporting

slandelle:
In the meantime, you can just set a 0 Content-Length header manually.

PR #3424, opened 2018-02-22
added import to documentation (assertions didn't work without it)

vitaliis:
an example didn't work without it. Actually none of the functions/scopes described were usable without this import.
This is an attempt to improve documentation, as I stumbled upon the problem while trying to do things described.

### DIFF
--- a/src/sphinx/general/assertions.rst
+++ b/src/sphinx/general/assertions.rst
@@ -27,7 +27,10 @@ All the assertions are evaluated after running the simulation. If at least one a
 Scope
 =====
 
-An assertion can test a statistic calculated from all requests or only a part.
+An assertion can test a statistic calculated from all requests or only a part. 
+To use a scope you need to add this import ``import io.gatling.core.Predef._``.
+
+Following scopes are available:
 
 * ``global``: use statistics calculated from all requests.
 


slandelle:
Hi.

Thanks for reporting. Actually, those imports should be there everywhere you want to use Gatling's DSL. There's a (too) quick mention in the [Quickstart](https://gatling.io/docs/current/quickstart/#gatling-scenario-explained).

IMHO, it would be worth mentioning this fact again at the beginning of the [Simulation Structure page](https://gatling.io/docs/current/general/simulation_structure/). But it would be very cumbersome and more complicate to repeat the same thing over and over again in every sample.

Would you be interested in providing such PR?

vitaliis:
You are right. I created this PR as a starting point for the discussion.
I might take a look at it on the weekend or after work.
It might make sense to add this section in the beginning of each documentation page, which has examples requiring those imports. For example I took a look at the examples on the next page. And they also require those imports. I wouldn't know that the imports I need are exactly on the "Simulation Structure" page.

Another idea I had is to add these 3 necessary imports in the top of the gatling cheatsheet.
Could you point me where I could propose changes to the cheatsheet? I would like to propose more changes there.

slandelle:
https://github.com/gatling/gatling/blob/master/src/sphinx/cheat-sheet.html

slandelle:
Actually, we'll have a section mentioning imports, see https://github.com/gatling/gatling/commit/81332b371aa962ff7036141aab4fa98cf495aa7b#diff-21eb74d55803f1dab79d8fc14be5545d.

Thanks anyway!

PR #3428, opened 2018-03-01
JMS: Tracker Destination can by specified independent from replyTo Destination with doc

opensource21:
Fix the comments from #3401, adjust the code, so that at can be merged into master and added documentation.
The motivation for the pull request can be found in detail at #3401.

### DIFF
--- a/src/sphinx/cheat-sheet.html
+++ b/src/sphinx/cheat-sheet.html
@@ -3118,6 +3118,21 @@
               </div>
 
               <div class="function">
+                <label class="function-syntax" for="trackerDestination">trackerDestination</label>
+                <input type="checkbox" id="trackerDestination" name="trackerDestination">
+
+                <div class="function-details">
+                  On default gatling tracks the replies on the <code>replyDestination</code>.
+                  Optionally you can overwrite this destination with the parameter <code>trackerDestination</code>
+                  <dl>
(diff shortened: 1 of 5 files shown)


src/sphinx/cheat-sheet.html
```diff
@@ -3048,7 +3048,7 @@ <h3>Start</h3>
               <h3>Commons</h3>
 
               <div class="function">
-                <label class="function-syntax" for="reqreply">reqreply</label>
+                <label class="function-syntax" for="reqreply">requestReply</label>
```
slandelle: Woops, nice catch, thanks!

slandelle:
Thanks!

Cherry-picked as 7b52740

opensource21:
Are there any descriptions how to build the download-zip file? Or exists a specific zip-file?

slandelle:
https://gatling.io/docs/current/developing_gatling/building_gatling/

opensource21:
Thanks for the information. Is it correct that the conf-folder is missing and zinc is no longer in a special folder in lib?

slandelle:
> Is it correct that the conf-folder is missing

Right, I think you need to run sbt package in gatling to have it generated

> zinc is no longer in a special folder in lib

yes

PR #3440, opened 2018-03-16
Tiny formatting change

kaneeec:
The `js` directory is highlighted to stand up more from the text. It's hard to notice and people may actually miss it, partially because `js` is a strange directory where to store statistics. More suitable would be the root or separate `stats` directory


slandelle:
Cherry-picked, thanks!

PR #3450, opened 2018-04-09
Doc: Fix typo in gatling-defaults.conf file

ssteeg-mdsol:
Change `hashshake` → `handshake` in the gatling-defaults.conf file


slandelle:
cherry-picked, thanks!

Issue #3445, opened 2018-03-23
Gatling 3 JMS performance  with IBM MQ8


LukasHavemann:
Hello,

we used the latest gatling 2 release for jms loadtesting with 8000 req/sec against IBM MQ 8 and had no problems. With gatling 3 we have big performance issues with just 4000 req/sec. 

To verify that gatling is the cause, we setup our gatling 2 & 3 loadtestdriver on one machine, a teststub on a second machine and IBM MQ 8 on a thrid machine. We ran the same scenario with gatling 2 and gatling 3 multiple times. Here are the two reports with the same machines, same teststub and same scenarios. 

Best Regards
Lukas

Gatling 2
![gatling2](https://user-images.githubusercontent.com/655563/37845054-caefef46-2ec9-11e8-93f4-023928c6b4af.PNG)

Gatling 3
![gatling3_performance](https://user-images.githubusercontent.com/655563/37845060-cf2f547a-2ec9-11e8-83f2-aa66e2fca25f.PNG)

slandelle:
We don't see such performance regression on our internal benchmarks, where we test Artemis.

Results below where produced on my laptop (I would expect way better results on our Linux boxes):
<img width="900" alt="artemis" src="https://user-images.githubusercontent.com/849197/37864354-b564418c-2f6d-11e8-9615-fba04b04ae55.png">

Several possibilities:

* The main difference between Gatling 2 and Gatling 3 JMS support is that we now use `javax.jms.MessageListener` instead of  blocking`javax.jms.MessageConsumer#receive`. If that's the issue with MQ, I advice you get in touch with your IBM support to get this fixed.
* The problem is with your platform and is not related with Gatling
* The problem is indeed with Gatling

If you want us to investigate the latter, please provide a complete standalone application that reproduces your issue and we can run on our side (hence not based on IBM MQ that we obviously don't have licence for).

LukasHavemann:
We do not see this behavior with low req/sec rates. With 1500 req/sec Gatling 3.0 performs nearly as good as Gatling 2.0.

According to MQExplorer it seems that there is only one consumer that receives the messages. So it seems that the IBM mQ provider starts only one receiver thread. This corresponds with the behavior we experienced in our own application. We measured that one session + one receiver thread was capable of receiving ~2400 messages per sec not more. It seems that IBM MQ needs multiple paralell sessions in order  to surpass the 2400 message treshhold. This was possible with Gatling 2.0, since the number of paralell consumers was configurable.

[...]

PR #3451, opened 2018-04-09
JMS-listener thread count can be configured again

LukasHavemann:
solves #3445

### DIFF
--- a/gatling-core/src/main/resources/gatling-defaults.conf
+++ b/gatling-core/src/main/resources/gatling-defaults.conf
@@ -105,7 +105,7 @@ gatling {
   }
   jms {
     replyTimeoutScanPeriod = 1000  # scan period for timedout reply messages
-    listenerThreadCount = 1        # listener thread count for receiveing and diaptching jms messages
+    listenerThreadCount = 1        # listener thread count for receiveing and dispatching jms messages
   }
   data {
     writers = [console, file]      # The list of DataWriters to which Gatling write simulation data (currently supported : console, file, graphite, jdbc)
--- a/gatling-core/src/main/resources/gatling-defaults.conf
+++ b/gatling-core/src/main/resources/gatling-defaults.conf
@@ -105,6 +105,7 @@ gatling {
   }
   jms {
     replyTimeoutScanPeriod = 1000  # scan period for timedout reply messages
+    listenerThreadCount = 1        # listener thread count for receiveing and diaptching jms messages
   }
   data {
     writers = [console, file]      # The list of DataWriters to which Gatling write simulation data (currently supported : console, file, graphite, jdbc)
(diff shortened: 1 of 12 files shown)


slandelle:
So if I get it right, your perf issue with IBM MQ is that it's slow on consuming messages and needs several consumers?

gatling-core/src/main/resources/gatling-defaults.conf
```diff
@@ -105,6 +105,7 @@ gatling {
   }
   jms {
     replyTimeoutScanPeriod = 1000  # scan period for timedout reply messages
+    listenerThreadCount = 1        # listener thread count for receiveing and dispatching jms messages
```
slandelle: I think such parameter would be more visible for users if passed from JmsProtocol instead of config.

LukasHavemann:
Yes. IBM MQ creates only one consumer. I tried to increase the client thread pool size with  `com.ibm.ws.wmqcsi.threadPoolMaximumSize=15` [1]. This helped a little bit, but overall performance was still bad, due two only one consumer. 

With this patch galing 3 + IBM mq performs as great as gatling 2 + IBM mq.

[1] https://www.ibm.com/support/knowledgecenter/en/SSAW57_7.0.0/com.ibm.websphere.nd.iseries.doc/info/iseriesnd/ae/tmj_adm34.html

gatling-jms/src/main/scala/io/gatling/jms/protocol/JmsProtocol.scala
```diff
@@ -46,11 +46,12 @@ object JmsProtocol {
...
-    connectionFactory: ConnectionFactory,
-    credentials:       Option[Credentials],
-    deliveryMode:      Int,
-    replyTimeout:      Option[Long],
-    messageMatcher:    JmsMessageMatcher
+    connectionFactory:    ConnectionFactory,
+    credentials:          Option[Credentials],
+    deliveryMode:         Int,
+    replyTimeout:         Option[Long],
+    listenerThreadCount:  Option[Int],
```
slandelle: Don't make it an `Option[Int]`, just make it an `Int`.

gatling-jms/src/main/scala/io/gatling/jms/protocol/JmsProtocolBuilder.scala
```diff
@@ -33,11 +33,12 @@ case object JmsProtocolBuilderBase {
...
-    connectionFactory: ConnectionFactory,
-    creds:             Option[Credentials] = None,
-    deliveryMode:      Int                 = DeliveryMode.NON_PERSISTENT,
-    messageMatcher:    JmsMessageMatcher   = MessageIDMessageMatcher,
-    replyTimeout:      Option[Long]        = None
+    connectionFactory:    ConnectionFactory,
+    creds:                Option[Credentials] = None,
+    deliveryMode:         Int                 = DeliveryMode.NON_PERSISTENT,
+    messageMatcher:       JmsMessageMatcher   = MessageIDMessageMatcher,
+    listenerThreadCount:  Option[Int]         = None,
```
slandelle: Don't make it an `Option[Int]`, make it an `Int` whit default value of 1.

slandelle:
cherry-picked, thanks!

PR #3463, opened 2018-05-17
Lookup JMS Queues and Topics via JNDI

sventorben:
JEE recommends to lookup JMS Queues and Topics via JNDI instead of using
createQueue/createTopic.

see https://docs.oracle.com/javaee/7/api/javax/jms/Session.html#createQueue-java.lang.String-

### DIFF
--- a/gatling-jms/src/main/scala/io/gatling/jms/client/JmsClient.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/client/JmsClient.scala
@@ -84,8 +84,8 @@ abstract class JmsClient(
 
   def createDestination(destination: JmsDestination): Destination =
     destination match {
-      case JmsQueue(name)    => session.createQueue(name)
-      case JmsTopic(name)    => session.createTopic(name)
+      case JmsQueue(name)    => ctx.lookup(name).asInstanceOf[Queue]
+      case JmsTopic(name)    => ctx.lookup(name).asInstanceOf[Topic]
       case JmsTemporaryQueue => session.createTemporaryQueue()
       case JmsTemporaryTopic => session.createTemporaryTopic()
     }


slandelle:
Thanks for contributing.
Is there any actual issue with using create? You might have realized that the patch you suggest doesn’t work as is on master/Gatling 3 where we support JMS providers which don’t support JNDI, such as AWS SQS.

sventorben:
I am facing an issue when using Gatling with RabbitMQ. There is an outdated plugin (see https://github.com/maiha/gatling-amqp). However, I could not get it to work with current versions of Gatling.
So, I thought of using the RabbitMQ JMS client (https://github.com/rabbitmq/rabbitmq-jms-client) which allows bridging from JMS to AMQP. This only works when using non-standard JMS API for configuring Queues and Topics. Therefore, it is recommended to configure the queues/topics an make them available via JNDI. 
Since the ConnectionFactory is looked up via JNDI, I thought this would be reasonable to lookup queues/topics the same way.

slandelle:
> This only works when using non-standard JMS API for configuring Queues and Topics. Therefore, it is recommended to configure the queues/topics an make them available via JNDI.

Could you please elaborate? [RMQSession seems to implement createQueue](https://github.com/rabbitmq/rabbitmq-jms-client/blob/master/src/main/java/com/rabbitmq/jms/client/RMQSession.java#L782).

> Since the ConnectionFactory is looked up via JNDI, I thought this would be reasonable to lookup queues/topics the same way.

Now ConnectionFactory can be directly created and some providers only support this way.

sventorben:
Naturally the RMQSession implements createQueue, because it is part of the JMS spec. [createQueue calls](https://github.com/rabbitmq/rabbitmq-jms-client/blob/v1.9.0/src/main/java/com/rabbitmq/jms/client/RMQSession.java#L782) `RMQDestination dest = new RMQDestination(queueName, true, false)`, but that constructor does not support AMQP compatibility. For AMQP compatibility RMQSession has an [additional constructor](https://github.com/rabbitmq/rabbitmq-jms-client/blob/v1.9.0/src/main/java/com/rabbitmq/jms/admin/RMQDestination.java#L94) 
`public RMQDestination(String destName, String amqpExchangeName, String amqpRoutingKey, String amqpQueueName)`

Of course, AMQP compatibility is not part of the JMS API. RabbitMQ JMS Client works around that problem by allowing to lookup AMQP compatible queues from JNDI. That feature is implemented in [RMQObjectFactory](https://github.com/rabbitmq/rabbitmq-jms-client/blob/v1.9.0/src/main/java/com/rabbitmq/jms/admin/RMQObjectFactory.java#L215) that implements [ObjectFactory](https://docs.oracle.com/javase/7/docs/api/javax/naming/spi/ObjectFactory.html). This allows JNDI framework to load object implementations dynamically.

When you are saying that other JMS providers do not support JNDI lookups, I assume you are referring to [Amazons SQS library](https://github.com/awslabs/amazon-sqs-java-messaging-lib), right?
From my point of view, libraries like SQS lib do not fully comply to the [JMS spec](http://download.oracle.com/otndocs/jcp/7195-jms-1.1-fr-spec-oth-JSpec/), which demands JNDI support:

> 1.4.6 Java Naming and Directory InterfaceTM (JNDI) API
> JMS clients look up configured JMS objects using the JNDI API. JMS
> administrators use provider-specific facilities for creating and configuring
> these objects.

and

> 2.6.1 Developing a JMS Client
> A typical JMS client executes the following JMS setup procedure:
> - Use JNDI to find a ConnectionFactory object
> - Use JNDI to find one or more Destination objects

and

> 9.1.2 Getting a Destination
> An administrator has created and configured a Queue named “StockSource”
> which is where stock quote messages are sent and received. Again, the
> destination can be looked up using the JNDI API.
> ```
> Queue stockQueue;
> stockQueue = (Queue)messaging.lookup("StockSource")
> ```

From my point of view it would be reasonable to change the behaviour as I proposed. In additional implementations like the AWS SQS lib should be fixed accordingly. This can easily be done by providing a ConnectionFactory and ObjectFactory, e.g. like it is done by the RabbitMQ client [over here](https://github.com/rabbitmq/rabbitmq-jms-client/tree/v1.9.0/src/main/java/com/rabbitmq/jms/admin).

PR #3484, opened 2018-07-02
doc compile error when sbt publishM2

jiangcaocn:
ERROR message:

[info] Main Scala API documentation to /pp/gatling/gatling-test-framework/target/api...
[error] /pp/gatling/gatling-core/src/main/java/io/gatling/core/jodd/FixedLazyMap.java:65:14: not found: type Entry
[error]   public Set<Entry<Object, Object>> entrySet() {
[error]

### DIFF
--- a/gatling-core/src/main/java/io/gatling/core/jodd/FixedLazyMap.java
+++ b/gatling-core/src/main/java/io/gatling/core/jodd/FixedLazyMap.java
@@ -62,7 +62,7 @@ public class FixedLazyMap extends AbstractMap {
   }
 
   @Override
-  public Set<Entry<Object, Object>> entrySet() {
+  public Set<Map.Entry<Object, Object>> entrySet() {
     buildIfNeeded();
     return map.entrySet();
   }


slandelle:
This is very weird:

* this class compiles just fine on my side and on Travis: https://travis-ci.org/gatling/gatling/builds/398800362
* `Entry` is directly visible from here, as `FixedLazyMap` implements `Map`
* The original class doesn't need to prefix either: https://github.com/oblac/jodd/blob/master/jodd-json/src/main/java/jodd/json/LazyMap.java#L91

Could you please explain how you got a failing build?

jiangcaocn:
I got the error when run "sbt publishM2", it's an error related to **compile of API document**

slandelle:
```
gatling-parent > clean
[success] Total time: 1 s, completed 2 juil. 2018 08:45:02
gatling-parent > publishM2
[success] Total time: 64 s, completed 2 juil. 2018 08:46:11
```
works just fine.

Which JDK do you use for compiling?

jiangcaocn:
I can reproduce the issue when try your steps: **clean** then **publishM2**.
I use:
java version "1.8.0_161"
Java(TM) SE Runtime Environment (build 1.8.0_161-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.161-b12, mixed mode)

slandelle:
Actually, I can see your error in the Travis logs, even though it's non blocking:

```
[error] /home/travis/build/gatling/gatling/gatling-core/src/main/java/io/gatling/core/jodd/FixedLazyMap.java:65:14: not found: type Entry
[error]   public Set<Entry<Object, Object>> entrySet() {
[error]              ^
```

The thing is this code compiles just fine and I suspect it's a bug in Scaladoc.

jiangcaocn:
Oh, yes, it didn't blocking the publish process. I agree we can leave the error and wait Scaladoc's fix.

slandelle:
Actually, I checked `java.util.HashMap` source code and it uses `Map.Entry`. I guess it's a good practice to prefix classes defined in an interface.
I've cherry-picked your commit, see a41832acbd33db977ff53fa197d9611a884af57f.

Thanks!

PR #3507, opened 2018-08-03
Update session_api.rst

WilliamPenado:
There are just two 'typos' in the lesson.
I added the word: 'be' where it states: "Using ``as`` will probably easier for most users". and
I deleted the second word: 'be' where it states: "(e.g. could be be properly chained with other action because a dot is missing)".


slandelle:
Thanks!

cherry-picked

PR #3508, opened 2018-08-10
Update sse.rst

kuskmen:
Doesn't it make sense first to show how to build a check and then explain how to set it? Also I see cancel check paragraph in the site, but I don't see it here, not sure if this is intended or I am missing something, also I must admit I am not quite familiar with how .rst work, so forgive me if I am talking nonsense and disregard. :)

### DIFF
--- a/src/sphinx/http/sse.rst
+++ b/src/sphinx/http/sse.rst
@@ -59,6 +59,20 @@ Gatling currently only support one check at a time per SSE stream.
 
 .. _http-sse-check-set:
 
+Build a Check
+-------------
+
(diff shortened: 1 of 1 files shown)


slandelle:
> Doesn't it make sense first to show how to build a check and then explain how to set it?

Makes sense

> Also I see cancel check paragraph in the site, but I don't see it here

Because what you see online is the latest released open source version (from the 2.3 branch) while master is for next gen 3.0 (that our FrontLine customers already use) where cancel is being dropped for now.

slandelle:
cherry-picked. thanks!

PR #3533, opened 2018-09-21
Pass simulation instance programmatically to Gatling

d2e8f9:
I want to integrate Gatling directly into Scalatest for performance and functional testing. Therefore a simulation must be passed programatically to gatling. 

Please see this question at Stackoverflow https://stackoverflow.com/questions/30194555/programmatically-execute-gatling-tests


slandelle:
We can accept to modify this code if it serves a purpose in another official Gatling component (that could be contributed too).
But we don’t want to open our internals for the sole sake of third parties and proprietary integrations.

d2e8f9:
Okay. Give me a little bit time. Maybe i come up with a more general component to integrate Gatling in ScalaTest for functional testing.

d2e8f9:
I want to withdraw this pull requests. Sorry for the inconvenience, but i was a little bit to fast. After inspecting more internals of gatling - the Predef singletons - ,  i see that only one simulation must run at the same time. Of course this makes perfectly sense for performance tests. Forking ensures this assumption in the gatling modules of sbt and maven. But in the tied integration with scalatest, i do not want to fork. So constructor parameters could be passed to a concrete simulation instance.  That's why a general tied integration is in my opinion to fragile and too custom (nested testsuite).

slandelle:
@d2e8f9 Yeah, I was thinking about what you were trying to do and I suspected it wouldn't work.
Gatling DSL requires several components to be available in scope, typically the configuration which is pulled from an implicit mutable reference from core.Predef. I know this looks ugly at first, but that's the most user friendly way, eg:

```scala

object SomeObject {
  // we need the resources location from the config so we can read the file
  val data: Seq[Map[String, String]] = csv("foo.csv").readRecords
}
```

I can't think of a better solution :(

d2e8f9:
Yeah, i see. Please take a look at this [playground](https://github.com/d2e8f9/gatling-scalatest-playground/tree/master/src/test/scala), where you can find a runnable solution for the integration as far as i know. There the simulation is created lazy after the GatlingConfiguration is set to the Predef variable. But this is not as robust as it should be for a general component. Nested suites ensures, that the simualtions are executed in serial. Each user must create such a GatlingSimulationSuite suite and i guess this creation is difficult to enforce in general.

PR #3555, opened 2018-10-08
add type anntation for better documentation

HenryDev:
Since this is used in https://gatling.io/docs/current/advanced_tutorial/#step-01-isolate-processes it would be better to label the type of the value, so that people would have a easier time to get it to compile.


slandelle:
Thanks for contributing, but those types are inferred and not explicit for a reason: they are not part of the public API (which is the DSL) so we can change them if needed.

PR #3562, opened 2018-10-13
Add request body dumping in scenario exporter, close #3561

### LINKED ISSUE #3561: Recorder: never dump request bodies
As of 3.0.0-RC4, the recorder never dumps the request bodies into a corresponding request.txt file. I looked into the code and the function "requestBodyFileName" in ScenarioExporter is only used for generating the name in the generated simulation file.

To reproduce, simply record a call to a REST api with a JSON body.

As of the comment in the code :

```scala
// dump request & response bodies if needed
requestElements.foreach(el => el.body.foreach {
  case RequestBodyBytes(bytes) => dumpBody(requestBodyFileName(el), bytes)
  case _                       =>
})
```

it should dump de request.

psychobaatezu:
Possible fix for issue#3561


slandelle:
Thanks, cherry-picked

PR #3578, opened 2018-10-22
Make parentheses-vs-curly-braces more explicit

asarturas:
In documentation the warnings about curly braces not working for randomSwitch methods are put in comments, which are practically invisible in most cases. I suggest to add a more explicit warning about that to avoid a confusion, as using braces does compile and breaks during runtime only.

### DIFF
--- a/src/sphinx/general/scenario.rst
+++ b/src/sphinx/general/scenario.rst
@@ -325,6 +325,8 @@ Similar to ``doSwitch``, but with a fallback if no switch is selected.
 ``randomSwitch`` can be used to emulate simple Markov chains.
 Simple means cyclic graphs are not currently supported.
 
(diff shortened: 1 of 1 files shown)


src/sphinx/general/scenario.rst
```diff
@@ -325,6 +325,8 @@ Similar to ``doSwitch``, but with a fallback if no switch is selected.
 ``randomSwitch`` can be used to emulate simple Markov chains.
 Simple means cyclic graphs are not currently supported.
 
+.. warning:: Beware: you have to use parentheses when using these switch methods, curly braces will change the semantic and will not work as expected.
```
slandelle: What do you mean by "change the semantic" and "no work as expected"? It's not like it would seem to work but actually do something different than expected. It simply won't compile.

PR #3579, opened 2018-10-22
Add ability to provide custom report generator

patelh:
Add ability to provide customer file/charts report generator through gatling config file.


slandelle:
Sorry, but `RunResultProcessor` and `ReportsGenerator` are not a public API and we have no plan to make them one atm.

PR #3582, opened 2018-10-23
Add ability to provide custom report generator

patelh:
Add ability to provide customer file/charts report generator through gatling config file.


patelh:
Gatling owners closed this PR request since it conflicts with their paid product Gatling Frontline.

See [here](https://github.com/gatling/gatling/pull/3579)

Makes gatling difficult to use outside simple testing without recording results and tracking them historically.

Keeping this PR in case someone wants to roll their own gatling version with this change.

PR #3588, opened 2018-10-29
Add throttling for JMS sending, close #3587

### LINKED ISSUE #3587: JMS: requests are not throttled on a global level
Tested against gatling 3.0.0. Same behaviour was in 2.3.0.

We have a simple scenario where we want to send JMS messages on to a queue in fire and forget style. We do not want to send more than 10 messages per second for one minute.
Given the scenario below gatling procudes as much messagea as it can onto the queue. The same code works with the http protocol without any issues:

```
setUp(
    scenario("jms test")
        .during(1 minute) {
          exec(
            jms("test")
              .send
              .queue("testqueue")
              .textMessage("test")
          )
        }
        .inject(atOnceUsers(1))
  )
    .throttle(reachRps(10) in (1 minute), holdFor(30 days))
    .protocols(jmsProtocol)
```

One work around seems to be to add an additional http request in the `during` block which does and fake http request to som empty endpoint so the throttle has an effect.

fkoehler:
This PR implements throttling with JMS

### DIFF
--- a/gatling-jms/src/main/scala/io/gatling/jms/action/JmsAction.scala
+++ b/gatling-jms/src/main/scala/io/gatling/jms/action/JmsAction.scala
@@ -20,13 +20,20 @@ import javax.jms.Message
 
 import io.gatling.commons.validation._
 import io.gatling.core.action.RequestAction
+import io.gatling.core.controller.throttle.Throttler
 import io.gatling.core.session._
 import io.gatling.core.util.NameGen
 import io.gatling.jms.client.{ JmsConnection, JmsConnectionPool }
 import io.gatling.jms.protocol.JmsProtocol
 import io.gatling.jms.request._
 
-abstract class JmsAction(attributes: JmsAttributes, protocol: JmsProtocol, pool: JmsConnectionPool)
+abstract class JmsAction(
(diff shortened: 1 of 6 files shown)


gatling-jms/src/main/scala/io/gatling/jms/action/RequestReply.scala
```diff
@@ -17,10 +17,10 @@
 package io.gatling.jms.action
 
 import io.gatling.commons.util.Clock
-
```
slandelle: please restore empty line

gatling-jms/src/main/scala/io/gatling/jms/action/JmsAction.scala
```diff
@@ -17,16 +17,22 @@
 package io.gatling.jms.action
 
 import javax.jms.Message
-
```
slandelle: please restore empty line

fkoehler:
I will change the mentioned things. How about tests?

slandelle:
I don't think it's worth adding tests for what you've changed (I mean, it's pretty complicated to test and the change is pretty straightforward).
It would be great to have tests on Throttler, though, but that's another matter.

PR #3596, opened 2018-10-31
Allow recorder to emit simulationClass instead of "request", close #3583

### LINKED ISSUE #3583: Recorder: Make "request_" prefix in recorder customizable / simulation dependent
The constant "request_" prefix in recordings / converted HARs should be customizable or dependent on the simulation name.

My usecase is the following:
I'm recording different cases in an application using the recorder and / or converting multiple HAR Files to separate Simulations. In addition to that I have a hand written Simulation that aggregates the scenarios from these automatically generated recordings. 

```scala
import io.gatling.http.Predef._

class MainSim extends Simulation {
  val httpProtocol = http
    .baseURL("https://localhost:8080")
    .contentTypeHeader("application/json")
    .acceptHeader("application/json, application/*+json")
    .acceptEncodingHeader("gzip,deflate")

  setUp(List(
    new userLoginCase().scn.inject(
      rampUsers(20) over (5 seconds),
      constantUsersPerSec(20) during (60 seconds)),
    new userSearchCase().scn.inject(
      rampUsers(20) over (5 seconds),
      constantUsersPerSec(20) during (60 seconds))
  )).protocols(httpProtocol)

}
```

I do not want to edit the automatically generated scala files as these are created by our test / integration pipeline. Instead I specify all the necessary information in my custom Simulation that aggregates the other Simulations.

The issue that I'm currently facing is that all of the recorded / converted simulations have requests named "request_0", "request_1" etc. Since the name of the request is the same these requests get merged in the simulation and report. In reality the content is quite different though and I'd like to track them separately.

One way to deal with this would be to introduce a way (command line argument) that allows me to specify a custom prefix for the recording / conversion for use instead of  "request_", for example "login_" and "search_" in the above example.

An alternative option would be to name the request after simulation class instead of using the fixed "request_" pattern. In the above example this might be "userLoginCase_request_0" for example. As this might be a breaking change for some users this would probably still require a way (e.g. an option) to opt in to that behavior.

cfraenkel:
Add a new config option `useSimulationAsPrefix` and corresponding functionality and tests. The default behavior is to still emit `request_` in the generated simulations.


slandelle:
> From what I've seen the request and response body RawFileBody calls already use the simulation name as a prefix. Where you thinking of those or different locations?

You're right, my bad.

cfraenkel:
I've added the setting to the UI and documentation. I decided to call it "Class Name"  instead of simulation class name so that it matches the Class Name field in the previous row in the UI.

I haven't tested the document generation as I do not currently have sphinx installed, but I matched the other sections in the file so I expect it will render correctly.

slandelle:
cherry-picked, thanks!

Note that I changed the Swing layout a bit to have options aligned.

Issue #3340, opened 2017-08-04
Support using another slf4j backend than logback


josetesan:
We are using gatling 2.2.5 as part of our maven build process, and as such, it does inheritance all the libraries our product uses : slf4j, log4j1.x, log4j2.x ( i know, i  know , but it's a legacy product can't be changed ).

The usage of logback inside gatling conflicts with these libraries, as there's a race condition between them, so we're giving an error :

```
SLF4J: Found binding in [jar:file:/D:/Users/OU83MP/.m2/repository/org/apache/logging/log4j/log4j-slf4j-impl/2.2/log4j-slf4j-impl-2.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/D:/Users/OU83MP/.m2/repository/org/slf4j/slf4j-log4j12/1.7.5/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/D:/Users/OU83MP/.m2/repository/ch/qos/logback/logback-classic/1.1.7/logback-classic-1.1.7.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory]
15:08:00.845 [main] TRACE io.gatling.app.Gatling$ - Starting
15:08:00.942 [main] TRACE io.gatling.app.Gatling$ - Configuration loaded
15:08:01.251 [GatlingSystem-akka.actor.default-dispatcher-4] INFO  akka.event.slf4j.Slf4jLogger - Slf4jLogger started
15:08:01.251 [main] TRACE io.gatling.app.Gatling$ - ActorSystem instantiated
15:08:01.266 [main] TRACE io.gatling.app.Gatling$ - Runner instantiated
15:08:01.358 [main] TRACE io.gatling.app.Runner - Running
java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #3605, opened 2018-11-14
Remove hard dependency on logback

sebastianvoss:
This will allow to choosing other logging frameworks especially when using the Gatling SBT plugin as part of an existing project.

Solves https://github.com/gatling/gatling/issues/3340

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -82,7 +82,7 @@ object Dependencies {
 
   // format: ON
 
-  private val loggingDeps = Seq(slf4jApi, scalaLogging, logback)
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -82,7 +82,7 @@ object Dependencies {
 
   // format: ON
 
-  private val loggingDeps = Seq(slf4jApi, scalaLogging, logback)
```
slandelle: Please revert. You're breaking all existing code and distributions.

slandelle:
@sebastianvoss One thing you can change though is the scope of the logback dep and make it "runtime" :)

slandelle:
thanks!

PR #3623, opened 2018-11-29
Removed incorrect log message when logging provider is log4j2

testphreak:
- Fail silently when stop method isn't invoked correctly


testphreak:
Fixes issue: https://github.com/gatling/gatling/issues/3624

slandelle:
@testphreak Thanks! I'll merge next week when I'll be back.

PR #3632, opened 2018-12-10
Update maven_plugin.rst

kolotyluk:
Clarification of maven lifecycle

### DIFF
--- a/src/sphinx/extensions/maven_plugin.rst
+++ b/src/sphinx/extensions/maven_plugin.rst
@@ -6,7 +6,7 @@ Maven plugin
 
 .. highlight:: xml
 
-Thanks to this plugin, Gatling can be launched when building your project, for example with your favorite CI solution.
+Using this plugin, Gatling can be launched when building your project, for example with your favorite Continuous Integration (CI) solution. Note: while you can invoke it explicitly using ``mvn gatling:test``, it cannot (at this time) be run implicitly as part of the normal maven lifecycle, for examole: ``mvn verify``.
 
 Versions
 ========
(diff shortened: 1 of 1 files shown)


slandelle:
> it cannot (at this time) be run implicitly as part of the normal maven lifecycle, for example: ``mvn verify``

This statement is wrong.

Only official maven plugins are automatically executed in the build lifecycle.
Community plugins have to be explicitly configured with `executions` blocks, see https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Using_the_executions_Tag.

src/sphinx/extensions/maven_plugin.rst
```diff
@@ -53,6 +53,30 @@ It's bound by default to the ``integration-test`` phase.
 
 .. _maven-advanced-configuration:
 
+Verify
+======
+
+To have Gatling run automatically during the integration-test phase,
+for example: ``mvn verify``, add something like the following to your pom.xml
```
slandelle: Actually, any phase from `integration-test` and after would trigger Gatling.

src/sphinx/extensions/maven_plugin.rst
```diff
@@ -53,6 +53,30 @@ It's bound by default to the ``integration-test`` phase.
...
+    <groupId>io.gatling</groupId>
+    <artifactId>gatling-maven-plugin</artifactId>
+    <version>MANUALLY_REPLACE_WITH_LATEST_VERSION</version>
+    <configuration>
+      <simulationClass>it.mySimulationIT</simulationClass>
+    </configuration>
+    <executions>
+      <execution>
+        <id>mySimulation</id>
+        <phase>integration-test</phase>
```
slandelle: `integration-test` is the default phase for the `test` goal so it can be omitted.

slandelle:
cherry-picked and changed some wording.

Thanks!

PR #3639, opened 2018-12-19
Change config priority

marcguilera:
Would it make sense to have properties passed in taking priority over the properties file? My rationale is that one might use the file as sane defaults that can be overridden with code.


slandelle:
What would make sense for you would’t make sense for others and we can’t definitively have a breaking change here, sorry.

Issue #3640, opened 2018-12-19
HTTP: Origin Header set incorrectly


jamesdbloom:
The Origin header is incorrect set to an invalid value on this line of code: https://github.com/gatling/gatling/blob/7477483964d07c932f3c509259aea9adbec7214c/gatling-http-client/src/main/java/io/gatling/http/client/RequestBuilder.java#L216 

Due this commit: https://github.com/gatling/gatling/commit/faf874ae6b55dc87fce8af33c8ba058c9d98a90a

The Origin header now has the value of the Host header with is an invalid value.  The HTTP spec says: `The Origin header is added by the user agent to describe the security contexts that caused the user agent to initiate an HTTP request. HTTP servers can use the Origin header to mitigate against Cross-Site Request Forgery (CSRF) vulnerabilities.`.

Given Gatling is executing API requests not from a browser or other agent host on a domain it is not valid to set the Origin header.

slandelle:
Suggested fix?

jamesdbloom:
I've submitted a PR with the suggested fix

jamesdbloom:
But the Origin header isn't a valid value for an API request, nor should it be set to the same value as the Host header.  This is inconsistent with both the HTTP specification and the common usage of this header, as per the section I quoted from the HTTP specification above.

slandelle:
Sorry for the long delay.

It turns out that the issue is more complex, see https://fetch.spec.whatwg.org/#origin-header.

Basically, we have 2 use cases:

1. regular requests whose method is neither `GET` nor `HEAD`: here, we should automatically set the `Origin` header as a pathless version of the `Referer` header (and not the target remote as it's currently done).
2. CORS: here it would probably be difficult to guess the proper value. I guess it leaves us with trusting the user to properly set it, most likely in the protocol config as a common header.

Here's what I suggest:
* add a `disableAutoOrigin` feature similar to `disableAutoReferer`, defaulting to `false`
* when `Origin` header is not set (in the request or the protocol) and `disableAutoOrigin` is not turned on and method is neither `GET` nor `HEAD` and `Referer` is header is defined, automatically add an `Origin` header computed as a pathless version of `Referer`

WDYT?

jamesdbloom:
I think what you suggest sounds good.  The only thing I am unsure of is why gatling needs to use CORS because its not making request from inside a browser.

slandelle:
> I think what you suggest sounds good.

Are you interested in contributing?

[...]

PR #3641, opened 2018-12-19
Removing invalid Origin header

jamesdbloom:
Fixes: #3640


slandelle:
Fixed in f8c99e5c9532d9808c38dbdc0506096a0c3b34d5

PR #3644, opened 2019-01-03
Implement ArrayBasedMap unsupported methods (+/-, updated), close #3625

### LINKED ISSUE #3625: Core: Implement ArrayBasedMap `+`, `-` and `updated` methods
Gatling 3.0.1

Use Case:
We read all feeder records (using https://gatling.io/docs/current/session/feeder/#grabbing-records) at the beginning of the simulation and make following modifications before virtual users start using them:
- Adding required fields if we find them missing
- Updating existing fields if we find them incorrect

Example:
```
val simDataTokens: Vector[Map[String, Any]] = {
  csv(dataFile).readRecords.as[Vector[Map[String, String]]]
    .zipWithIndex
    .map(r => {
      val id = r._2
      val tokens = r._1
      tokens + ("id" -> id) // or tokens.updated("id", id)
    })
}
```

Feature Request:
ArrayBasedMap methods for `+`, `-` and `updated` need to be implemented. Currently, they return java.lang.UnsupportedOperationException when called.

vu-illumio:
Stephane -

Please do not merge this untested PR. It still has compilation errors. I've done some readings about Scala types, but am still not able to resolve those errors.

At this time, I'd like to request your guidance for the proposed implementations:
1. Am I heading down the right path? Am I missing anything?
2. How do I resolve the compilation errors "No ClassTag available for K"?

Thank you.

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/feeder/ArrayBasedMap.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/feeder/ArrayBasedMap.scala
@@ -25,9 +25,9 @@ private[feeder] object ArrayBasedMap {
 
 private[feeder] class ArrayBasedMap[K, +V](keys: Array[K], values: Array[V], override val size: Int) extends AbstractMap[K, V] with Map[K, V] with Serializable {
 
-  override def +[V1 >: V](kv: (K, V1)) = throw new UnsupportedOperationException
+  override def +[V1 >: V](kv: (K, V1)): Map[K, V1] = updated(kv._1, kv._2)
 
-  override def updated[V1 >: V](key: K, value: V1): Map[K, V1] = throw new UnsupportedOperationException
+  override def updated[V1 >: V](key: K, value: V1): Map[K, V1] = Map.empty[K, V1] ++ this + (key -> value)
 
   override def get(key: K): Option[V] = {
     var i = 0
@@ -54,5 +54,5 @@ private[feeder] class ArrayBasedMap[K, +V](keys: Array[K], values: Array[V], ove
     }
   }
 
-  override def -(key: K) = throw new UnsupportedOperationException
+  override def -(key: K): Map[K, V] = Map.empty[K, V] ++ this - key
 }
--- a/gatling-core/src/test/scala/io/gatling/core/feeder/ArrayBasedMapSpec.scala
+++ b/gatling-core/src/test/scala/io/gatling/core/feeder/ArrayBasedMapSpec.scala
@@ -35,4 +35,28 @@ class ArrayBasedMapSpec extends BaseSpec {
   it should "return an empty iterator when there's no value" in {
     ArrayBasedMap(Array("col1", "col2"), Array.empty).iterator shouldBe empty
   }
+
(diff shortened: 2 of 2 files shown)


slandelle:
You don't have to return `ArrayBasedMap` instances, `Map` ones suffice.
Why don't you just add `this` into `Map.empty` so you'd get a default Scala implementation and then perform the operation on the result?

gatling-core/src/main/scala/io/gatling/core/feeder/ArrayBasedMap.scala
```diff
@@ -25,9 +25,17 @@ private[feeder] object ArrayBasedMap {
 
 private[feeder] class ArrayBasedMap[K, +V](keys: Array[K], values: Array[V], override val size: Int) extends AbstractMap[K, V] with Map[K, V] with Serializable {
 
-  override def +[V1 >: V](kv: (K, V1)) = throw new UnsupportedOperationException
+  override def +[V1 >: V](kv: (K, V1)): Map[K, V1] = updated(kv._1, kv._2)
 
-  override def updated[V1 >: V](key: K, value: V1): Map[K, V1] = throw new UnsupportedOperationException
+  override def updated[V1 >: V](key: K, value: V1): Map[K, V1] = {
```
slandelle: `Map.empty[K, V1] ++ this + (key -> value)` would do exactly the same thing :)

gatling-core/src/main/scala/io/gatling/core/feeder/ArrayBasedMap.scala
```diff
@@ -54,5 +62,15 @@ private[feeder] class ArrayBasedMap[K, +V](keys: Array[K], values: Array[V], ove
     }
   }
 
-  override def -(key: K) = throw new UnsupportedOperationException
+  override def -(key: K): Map[K, V] = {
```
slandelle: `Map.empty[K, V1] ++ this - key` is pretty similar (it might just add key and then remove it. I don't think we really care about performance here (or we'd have to benchmark) so I'd say we should go with the more simple form.

gatling-core/src/main/scala/io/gatling/core/feeder/ArrayBasedMap.scala
slandelle: :)

gatling-core/src/main/scala/io/gatling/core/feeder/ArrayBasedMap.scala
slandelle: Thanks

slandelle:
@vu-illumio Could you please squash your commits into a single one and change your commit message to match [our conventions](https://netty.io/wiki/writing-a-commit-message.html), please?

vu-illumio:
@slandelle - Please review commit [8f4507ec0d0cd06df0e88d0e55a8a2b23b8c05ff](https://github.com/gatling/gatling/pull/3644/commits/8f4507ec0d0cd06df0e88d0e55a8a2b23b8c05ff). Thanks.

slandelle:
Cherry-picked, thanks!

Issue #3648, opened 2019-01-10
HTTP: Set HTTP header content-length:0 for POST,PUT,PATCH methods when no body is send doesn't work


piotrbo:
Gatling 3.0.2 does not apply HTTP header content-length: 0 for POST,PUT,PATCH methods when no body is send.

This happens only when HTTP method is passed to io.gatling.http.request.builder.Http.httpRequest()
as String e.g. "POST" instead of HttpMethod.POST

`scenario("test").exec(http("empty POST test").httpRequest("POST","https://gatling.io/"))`

Gatling log shows no content-length header:

> =========================
HTTP request:
POST https://gatling.io/
headers=
accept: */*
origin: https://gatling.io
host: gatling.io
=========================

PR #3649, opened 2019-01-10
Fix workaround for servers that requires HTTP header content-length:0…

piotrbo:
… for POST,PUT,PATCH methods when no body is send. Closes gatling/gatling#3648

Motivation:

Some POST requests without body results with HTTP 411 Length Required
 error on Microsoft Azure cloud servers

Modification:

- httpRequest() passes now existing HttpMethod object instead of creating a new one
- In buildRequestWithoutBody() replace object reference comparison with equals() from comparable HttpMethod class

Result:

HTTP request header 'content-length: 0'  is set as intended for POST,PUT,PATCH methods when no body is send

### DIFF
--- a/gatling-http-client/src/main/java/io/gatling/http/client/impl/request/WritableRequestBuilder.java
+++ b/gatling-http-client/src/main/java/io/gatling/http/client/impl/request/WritableRequestBuilder.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 
 import static io.netty.handler.codec.http.HttpHeaderNames.*;
 import static io.netty.handler.codec.http.HttpMethod.*;
+import static java.util.Arrays.asList;
 
 public class WritableRequestBuilder {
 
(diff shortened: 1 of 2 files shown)


gatling-http-client/src/main/java/io/gatling/http/client/impl/request/WritableRequestBuilder.java
```diff
@@ -41,7 +42,7 @@ private static WritableRequest buildRequestWithoutBody(String url,
                                                          HttpHeaders headers) {
 
     // force content-length to 0 when method usually takes a body, some servers might break otherwise
-    if (!headers.contains(CONTENT_LENGTH) && (method == POST || method == PUT || method == PATCH)) {
+    if (!headers.contains(CONTENT_LENGTH) && asList(POST, PUT, PATCH).contains(method)) {
```
slandelle: Got bitten by this class being in Java and not Scala so == doesn’t mean equals...
Then, your suggested fix allocates a new ArrayList on each execution and then performs an array scan.
Please create a constant Set.

slandelle:
Thanks! Please address comment and ping again.

slandelle:
cherry-picked

Thanks!

PR #3668, opened 2019-02-06
Doc: data and bodies are now in resources dir

guikcd:
According to https://gatling.io/docs/current/migration_guides/2.3-to-3.0/
and https://gatling.io/docs/current/session/feeder/


slandelle:
Thanks!

cherry-picked

PR #3690, opened 2019-03-08
(Inactive) NATS Plugin

laugimethods:


slandelle:
cherry-picked, thanks!

PR #3700, opened 2019-04-07
enable disabled ConsoleDataWriterSpec

szymonprz:
Enabling disabled specs in ConsoleDataWriterSpec because they all pass and there is no reason to leave them disabled, and there was no reason written why they were disabled.


slandelle:
cherry-picked, thanks!

Issue #3706, opened 2019-04-11
Recorder: resource files are always dumped to root possibly overwriting files from a different class


pmalic:
Resource files (*request.txt/*response.txt) are always dumped to root of resources folder even when you supply a package name in Recorder. The resource files only contain the class name in their names so you can't have two classes with the same name in different packages because the resource files would be overwritten.

Wouldn't it make more sense to dump them into a subfolder named after the package name, just as is the case with .scala file?

I can of course manually create a subfolder under resources folder, move the resource files into it and change the paths in the .scala file, but this happening automatically would be great and would prevent someone mistakenly overwriting their resource files.

slandelle:
Looks like a good idea.
Willing to contribute?

pmalic:
This is in gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioExporter.scala, right?

Basically, transform requestBodyFileName and responseBodyFileName to be like simulationFilePath?

I haven't coded anything in Scala so far, but this seems fairly simple, I'd like to try :)

pmalic:
Fixed by 7652b6ad8e661e8aef102f405b32c11a58d88b2c

PR #3707, opened 2019-04-12
Recorder: dump resources to a subfolder named after package

pmalic:
Implemention for issue #3706.

master couldn't compile (using OpenJDK 11.0.2), had to change JdkXmlParsers.scala first (commit pmalic@96d6929).

Also changed string interpolators for `requestBodyFileName`/`responseBodyFileName` to `s`, it was `f` although there were no printf-style format strings.

### DIFF
--- a/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioExporter.scala
+++ b/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioExporter.scala
@@ -39,21 +39,35 @@ private[recorder] object ScenarioExporter extends StrictLogging {
 
   private val EventsGrouping = 100
 
+  private def packageAsFolderPath(separator: String)(implicit config: RecorderConfiguration) =
+    config.core.pkg.replace(".", separator)
+
   def simulationFilePath(implicit config: RecorderConfiguration): Path = {
     def getSimulationFileName: String = s"${config.core.className}.scala"
     def getSimulationsFolder = {
-      val path = config.core.simulationsFolder + File.separator + config.core.pkg.replace(".", File.separator)
+      val path = config.core.simulationsFolder + File.separator + packageAsFolderPath(File.separator)
       getFolder(path)
     }
 
     getSimulationsFolder / getSimulationFileName
   }
 
-  def requestBodyFileName(request: RequestElement)(implicit config: RecorderConfiguration) =
-    f"${config.core.className}_${request.id.leftPad(4, '0')}_request.txt"
+  private def resourcesFolder(implicit config: RecorderConfiguration): Path = {
+    val path = config.core.resourcesFolder + File.separator + packageAsFolderPath(File.separator) + File.separator + config.core.className
+    getFolder(path)
+  }
 
-  def responseBodyFileName(request: RequestElement)(implicit config: RecorderConfiguration) =
-    f"${config.core.className}_${request.id.leftPad(4, '0')}_response.txt"
+  private def requestBodyFileName(request: RequestElement) =
+    s"${request.id.leftPad(4, '0')}_request.txt"
+
+  def requestBodyRelativeFilePath(request: RequestElement)(implicit config: RecorderConfiguration) =
+    packageAsFolderPath("/") + "/" + config.core.className + "/" + requestBodyFileName(request)
+
+  private def responseBodyFileName(request: RequestElement) =
+    s"${request.id.leftPad(4, '0')}_response.txt"
+
+  def responseBodyRelativeFilePath(request: RequestElement)(implicit config: RecorderConfiguration) =
+    packageAsFolderPath("/") + "/" + config.core.className + "/" + responseBodyFileName(request)
 
   def exportScenario(harFilePath: String)(implicit config: RecorderConfiguration): Validation[Unit] =
     safely(error => s"Error while processing HAR file: $error") {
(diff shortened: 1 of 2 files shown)


pmalic:
Simulation files now have a subfolder referenced in RawFileBody(), for example:
```scala
.body(RawFileBody("pkg3/cls3_0012_request.txt"))
.check(bodyBytes.is(RawFileBody("pkg3/cls3_0012_response.txt")))
```

The separator in the string argument is hardcoded to `/`, `File.separator` is not used here because in this case a simulation recorded on Windows wouldn't work on Unixes, and `/` works on Windows. There are operating systems where `/` wouldn't work (e.g. OpenVMS), not sure would this be an issue for Gatling. 

I've considered other options, like making the simulation "runner" (don't know the correct term, looking into Gatling code for the first time) read from the subfolder by default, but this would break compatibility, people would have to move their old resources to subfolders. Another option I've considered is generating some code in the simulation file that would make the simulation "runner" look for resources in a subfolder, but to prevent again hardcoding the subfolder path, it would have to use reflection to figure out the package. This is beyond my current very limited Scala skills :)

slandelle:
> master couldn't compile (using OpenJDK 11.0.2), had to change JdkXmlParsers.scala first (commit pmalic/gatling@96d6929).

Gatling has to be compiled with JDK8. Compiling with JDK11 would produce bytecode that can't run on JDK8 (typically, because of `ByteBuffer#flip` that would cause LinkageError).
So this change is a no go.

> The separator in the string argument is hardcoded to / ..., not sure would this be an issue for Gatling.

It's OK because we mostly use classloader resolution.

gatling-core/src/main/scala/io/gatling/core/check/extractor/xpath/JdkXmlParsers.scala
```diff
@@ -74,7 +74,7 @@ class JdkXmlParsers(configuration: GatlingConfiguration) {
 
         override def getPrefix(uri: String): String = throw new UnsupportedOperationException
 
-        override def getPrefixes(uri: String): java.util.Iterator[_] = throw new UnsupportedOperationException
+        override def getPrefixes(uri: String): java.util.Iterator[String] = throw new UnsupportedOperationException
```
slandelle: Please revert

slandelle:
Moreover, could you please sign our contributor license agreement? https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw

gatling-core/src/main/scala/io/gatling/core/check/extractor/xpath/JdkXmlParsers.scala
pmalic: Reverted

pmalic:
Signed contributor license agreement.

slandelle:
Thanks!

Actually, I feel like only naming after the package is not enough.
We probably should turn the classname into a folder too (lower case), eg:

`foo.bar.BazSimulation` => `foo/bar/bazsimulation`

WDYT?

pmalic:
You already have resource file names prefixed with the class name, so you won't get a "conflict" with multiple classes in one package. The benefit would be to not have possibly thousands of files in one folder if you have multiple classes in a package and each class with a lot of resource files. Yeah, I think it makes sense to introduce another subfolder named after the class.

slandelle:
> Yeah, I think it makes sense to introduce another subfolder named after the class. Then the class name prefix in the resource file names should be dropped?

Yes, please.

pmalic:
Ok, it's now:
```scala
.body(RawFileBody("com/example/MySim/0005_request.txt"))
.check(bodyBytes.is(RawFileBody("com/example/MySim/0005_response.txt")))
```

pmalic:
I've also now implemented changes that set correct file extension to resource files so it's not always `.txt`. This makes opening them a lot easier. I extract mime type from `Content-Type` header. I have a map for a dozen most common mime types and for the rest `jodd.net.MimeTypes.findExtensionsByMimeTypes()` is used (it's based on Apache's `mime.types` file). I have a local map because this `MimeTypes` is not optimized for looking up extension from mime type but vice-versa (looked at the source code, they have a `HashMap` from extension to mime type). The changes touch the code I've already changed in this pull request and they are kind of related to this. Should I commit the changes to my branch associated with this pull request or should I make another one?

slandelle:
@pmalic No big deal, but those last changes regarding file extensions should have gone into a separate PR.

We're in the process of releasing, so we'll only be able to merge after that, all the more as I'm a bit concerned with the file path change on Windows that will require proper testing.

Thanks for your patience.

pmalic:
I didn't commit the changes related to file extensions to the branch associated with this PR, it's in a separate branch, I just pasted the link to the commit.

slandelle:
Ah, ok ,thanks!

pmalic:
Created PR #3711 for file extensions.

P.S. You mentioned testing on Windows, I'm mostly running simulations on Windows (I build gatling on macOS, but the app I'm load testing is on Windows).

slandelle:
@Neoware can you have a look please?

Neoware:
Hey @pmalic, this looks good except for the semicolon at the end of L56 :)

pmalic:
Thanks @Neoware for the review, will fix. I'll also squash the current commit with the fix, OK?

Neoware:
yes please :)

pmalic:
Done.

Neoware:
Your commit is merged :) Thank you for you contribution! Let's jump to the next one now :)

PR #3710, opened 2019-04-14
Improved assertion doc for websockets

leozilla:
Make clear that check name should be used when performing the `details` assertion for Websockets


slandelle:
Cherry-picked, thanks!

PR #3711, opened 2019-04-15
Recorder: Save resource files with correct file extension

pmalic:
Implemented changes that set correct file extension to resource files so it's not always `.txt`. This makes opening them a lot easier. Mime type is extracted from `Content-Type` header. There's a map for a dozen most common mime types and for the rest `jodd.net.MimeTypes.findExtensionsByMimeTypes()` is used (it's based on Apache's `mime.types` file). This is because `jodd.net.MimeTypes` is not optimized for looking up extension from mime type but vice-versa (looked at the source code, they have a `HashMap` from extension to mime type).

(Changes done here are based on changes already done in #3707, i.e. commit pmalic@378778dc84cf73dcd5642ce3c2039c468e7a3f03)

### DIFF
--- a/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioElement.scala
+++ b/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/ScenarioElement.scala
@@ -32,7 +32,9 @@ import io.gatling.recorder.config.RecorderConfiguration
 import io.gatling.recorder.model._
 import io.gatling.http.fetch.{ UserAgent => UserAgentHelper }
 
-import io.netty.handler.codec.http.{ DefaultHttpHeaders, HttpHeaders }
+import io.netty.handler.codec.http.{ DefaultHttpHeaders, HttpHeaders, HttpUtil }
+
+import jodd.net.MimeTypes
 
 private[recorder] case class TimedScenarioElement[+T <: ScenarioElement](sendTime: Long, arrivalTime: Long, element: T)
 
@@ -97,7 +99,7 @@ private[recorder] object RequestElement {
         requestHeaders
       }
 
-    RequestElement(request.uri, request.method, filteredRequestHeaders, requestBody, responseBody, response.status, embeddedResources)
+    RequestElement(request.uri, request.method, filteredRequestHeaders, requestBody, response.headers, responseBody, response.status, embeddedResources)
   }
 }
 
@@ -106,6 +108,7 @@ private[recorder] case class RequestElement(
     method:               String,
     headers:              HttpHeaders,
     body:                 Option[RequestBody],
+    responseHeaders:      HttpHeaders,
     responseBody:         Option[ResponseBody],
     statusCode:           Int,
     embeddedResources:    List[ConcurrentResource],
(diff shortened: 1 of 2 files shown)


slandelle:
@Neoware can you have a look please?

Neoware:
I think that the optimization with the map won't be useful, the simulation generation is not critical performance-wise. I would rather not have to maintain this map and to "duplicate" the logic.

Neoware:
Furthermore, you can simplify the getMimeType method using one option and map the function to perform HttpUtil.getMimeType only when the Option is a Some, and then using getOrElse to fallback on ApplicationOctetStream

Neoware:
the call to toString could be also move to getMimeType

Neoware:
You can now drop the first commit and rebase from master

pmalic:
Thanks @Neoware for the review.

Regarding `getMimeType`, you mean this:
```scala
def getMimeType(headers: HttpHeaders) = 
 Option(headers.get(ContentType)).map(HttpUtil.getMimeType).getOrElse(ApplicationOctetStream).toString
```
`getOrElse` will "pick-up" both `None` from `map(HttpUtil.getMimeType)` and possible `null` returned from `HttpUtil.getMimeType()`? (forgive me, I'm writing in Scala for the first time :)

P.S. Regarding the local map, there's another reason behind it I forgot to mention :) This `jodd.net.MimeTypes.findExtensionsByMimeTypes()` returns, for example, `list` file extension for `text/plain`, because for `text/plain` there are multiple extensions listed (`txt` being the first one) and `HashMap` is used so it's not necessarily returning the first one. This would work correctly if `LinkedHashMap` was used instead. I'll make a PR to jodd :)

Neoware:
yes you are right you will need to wrap HttpUtil.getMimeType() inside an Option, and to use flatMap instead of map :)

Neoware:
We have a mime type file that is used here https://github.com/gatling/gatling/blob/0a12b200425ce169285f0720f4d08c2c836e094a/gatling-http-client/src/main/java/io/gatling/http/client/body/multipart/FileLikePart.java#L30

Maybe we will do things with this instead of jodd. WDYT @slandelle ?

pmalic:
@Neoware please check 7953fe6760559047310c3fac6f338a252e6b24f3, if now OK, I'll squash.

pmalic:
FYI: I've tried with fixed `jodd.net.MimeTypes` and it then works correctly, i.e. `text/plain` gets `.txt` and not `.list`. Created oblac/jodd#701.

pmalic:
FYI: fix for `jodd.net.MimeTypes` got merged into master yesterday, I guess we'll have to wait for a release.

Neoware:
sorry for the delay! Yes I think we will wait for the jodd release to merge this :) and you can squash your 2 commits, it looks okay to me

pmalic:
Ok, squashed.

Neoware:
Jodd released a new version. Could you fix the tests failing before I merged this PR ?

slandelle:
cherry-picked

slandelle:
Thanks @pmalic and @Neoware

Neoware:
Your commit has been merged ! Thanks again for your two contributions :) Don't hesitate to suggest and contribute more improvements :)

pmalic:
Great :) 

Regarding further improvements, I have the following idea:

Ability to include request & response bodies "inline" in simulation source if their content type is text based (everything text/*) and their length is smaller than some number configurable in Recorder settings (by default it wouldn't "inline" anything, i.e. current behavior). It would take into account the character set of the bodies (e.g. "foo".getBytes("iso-8859-1")) and of the .scala file (I guess that's that setting in Recorder GUI).

Right now, if you are recording a "conversation" between a web app and some backend (i.e. basically API calls), you get a lot of small textual files that could easily be shown inline in the .scala file. This would increase readability of the source as you don't have go to the resources and open the files to see what is what.

WDYT? @Neoware @slandelle

Neoware:
That's an interesting idea, could you create a new issue so we can discuss about that new subject ?

pmalic:
Sure, here it is: #3731.

PR #3717, opened 2019-04-23
Update simulation_setup.rst

houchj:
looks like a typo


notdryft:
A typo it is indeed, thanks for pointing it out and for providing us with a PR.

FYI, "scaling out" would be better.

slandelle:
cherry-picked, thanks!

PR #3719, opened 2019-04-25
Update validation.rst, typo?

houchj:
also looks like a typo


notdryft:
Scalaz is a library dedicated to functional programming: https://github.com/scalaz/scalaz

Still, I understand the confusion that might occur for people who do not know about it. What about a link instead of a plain "Scalaz"?

slandelle:
Or just remove this mention about Scalaz.

PR #3720, opened 2019-04-29
Use the correct request name in the assertion documentation

leozilla:
Just a minor documentation fix for a part of the docs which I contributed a couple of weeks ago.

See: https://gatling.io/docs/current/general/assertions
The request name in the text does not match the code example

### DIFF
--- a/src/sphinx/general/assertions.rst
+++ b/src/sphinx/general/assertions.rst
@@ -35,7 +35,7 @@ An assertion can test a statistic calculated from all requests or only a part.
 
 * ``details(path)``: use statistics calculated from a group or a request. The path is defined like a Unix filesystem path.
 
-For example, to perform an assertion on the request ``Search``, use:
+For example, to perform an assertion on the request ``Index``, use:
 
 .. includecode:: code/AssertionSample.scala#details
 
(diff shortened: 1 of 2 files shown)


Neoware:
Hey! Thanks for spotting that error. 

Your PR doesn't concern the websocket part of the assertion documentation. 

Furthermore I think that the naming in this part of the documentation is very confusing. "Search" is considered a group and a request depending of the context, and the same name is used for the "Index" request inside and outside the group. If it was the case in a simulation, that would work, but I don't think that's the best way of explaining that feature. Maybe we could just change the name to something like:
- Search as a group, and Index as a request inside that group.
- Home as a request outside of a group.

Is it possible for you to fix that ?

If you are okay with these modifications, don't forget to also change the https://gatling.io/docs/current/general/assertions/#putting-it-all-together part of the documentation! 

Thanks in advance!

Neoware:
@leozilla do you agree on these suggestions ?

leozilla:
@Neoware yes I agree, sorry I totally forgot to adjust it, will do it this week

leozilla:
@Neoware I updated the PR, I hope I understood and implemented your suggestions correctly

slandelle:
@Neoware PTAL

Neoware:
hey, sorry for the delay !

Could you add a setUp example with the Home request in the Putting it all together?

Once it's finished could you squash your commits in a single one ?

Thanks in advance ! :)

slandelle:
Sorry for the very long wait. I've cleaned up the names, thanks for reporting!

PR #3762, opened 2019-07-11
Radius plugin listing in the community extensions

bpabdelkader:
Please fell free to list the [gatling-radius] plugin as active gatling extensions(https://github.com/bpabdelkader/gatling-radius)


slandelle:
@bpabdelkader  Sorry but your PR is broken.

PR #3774, opened 2019-07-29
use https to download build plugins from bintray

lbilger:
I noticed this because I am behind a proxy that blocks http connections. For a motivation to use https also see this article: https://medium.com/bugbountywriteup/want-to-take-over-the-java-ecosystem-all-you-need-is-a-mitm-1fc329d898fb.


slandelle:
cherry-picked

Thanks!

PR #3775, opened 2019-07-29
make graphite data writer flush one last time when simulation stops

lbilger:
The `GraphiteDataWriter` collects metrics and flushes them in a configurable interval. When the simulation is finished, there might be some data left over that is not flushed, especially if the interval is set to a value longer then the default of one second. This PR should make sure that the remaining data is flushed, similar to what the `ConsoleDataWriter` does.


slandelle:
cherry-picked

Thanks!

lbilger:
Thanks! Any idea when this will be released?

slandelle:
No ETA yet, as [there isn't much in the milestone atm](https://github.com/gatling/gatling/milestone/89?closed=1). Not before September at least, as we've just release 3.2.0.

PR #3776, opened 2019-07-29
update documentation to show the new writePeriod property instead of …

lbilger:
…the old writeInterval

It seems the property was renamed but not updated in the docs.


slandelle:
cherry-picked

Thanks!

PR #3797, opened 2019-09-13
converting the domain to an expression using el. close #3792

### LINKED ISSUE #3792: HTTP: getCookieValue withDomain does not accept an EL expression
My use case is the ability to test multiple URL's that actually resolve to the same multi-tenant application. I created a feeder that supplies a username and password as well as a baseUrl. 

The problem is that I want to extract a Cookie value to post in subsequent requests (XSRF protection) but I can't pass the value from the feed to the withDomain method.

```scala
    var login = exec(
        http("Authenticated")
          .get("${url}/api/v1/users/authenticated").asJson
          .check(status.is(403))
      )
      .exec(
        getCookieValue(
          CookieKey("XSRF-TOKEN")
            .withSecure(true)
            .saveAs("XSRF")
            .withDomain("${url}") <-- DOESN'T WORK
        )
      )

```

The error message I see is 

``` No Cookie matching parameters domain=${url}, path=/, name=XSRF-TOKEN, secure=true```

Is this something that can be fixed please

idanmid15:


slandelle:
Thanks, but that's not the correct way. GetCookieValueBuilder#domain should be an Option[Expression[String]] so conversion happens upstream.

slandelle:
LGTM now. Could you please sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw) please?

idanmid15:
Sure thing, just signed

slandelle:
Thanks!

PR #3821, opened 2019-11-06
Increment key IDs in doSwitch example code.

bradkarels:
Not related to any known issue, documentation only - no tests.


slandelle:
Thanks a lot!

Issue #3840, opened 2019-12-12
Upgrade moment js library from 2.8.3 to 2.24.0


clarkead:
Currently gatling-charts includes moment.js 2.8.3 which has two known medium severity security issues and one low severity issue (https://snyk.io/test/npm/moment/2.8.3). Upgrading to moment.js 2.24.0 remediates these issues and prevents the package from getting flagged as a security issue in builds.

PR #3841, opened 2019-12-12
#3840 - moment.js upgrade from 2.8.3 to 3.4.1 to address security issues.

clarkead:
Upgraded moment.js to address security issues outlined in ticket #3840

Tested load test report page with the new moment.js and ran unit tests and did not see any new warnings or errors.


slandelle:
Thanks a lot for contributing!

We'll upgrade thanks to your feedback, but we can't merge your PR as we can't ship a moment.js copy that comes from an untrusted source.

Thanks for your help and your understanding.

clarkead:
Thank you for the quick review and resolution.

PR #3850, opened 2020-01-07
Update jms.rst

jbertram:
JMS is not a protocol. It is an API.


slandelle:
Of course it is.
But then, all the load test tools that came before Gatling have been using for ages the word "protocol" for the communication thingies that's they support for generating load against a remote application, for lack of a better word. As a consequence, so do we. Here, we mean "our component that let you generate load with JMS".

jbertram:
I see now. Thanks for the clarification!

PR #3851, opened 2020-01-12
Update documentation sample gatling.conf

ghaxx:
I've found that "graphite" section should be inside "data". Also putting it inside "gatling" right in the example makes it less confusing at least to me.


slandelle:
cherry-picked

Thanks!

ghaxx:
You're welcome

PR #3864, opened 2020-02-09
Changed Cookie name to expression, close #3857

### LINKED ISSUE #3857: HTTP: Cannot apply Gatling EL to Cookie Name
I am using Gatling to load test one of our endpoints as we analyze throughput characteristics.  We are using a feeder to model user data and user IDs.  The system under test accepts cookie names and some of them are suffixed with a userID.

For example, when we try:

```
userId,token,
12345,MYSPECIALTOKEN
```
```
addCookie(Cookie("x-md-${userId}", "${token}"))
```

It appears that while I can use expressions for the Cookie Value, the Cookie Name cannot.  We see that the cookie `x-md-${userId}: MYSPECIALTOKEN` where we actually expected:
`x-md-12345: MYSPECIALTOKEN`

**ref**: https://github.com/gatling/gatling/blob/master/gatling-http/src/main/scala/io/gatling/http/action/cookie/AddCookieDsl.scala#L22

laaksonenl:


slandelle:
Hi @laaksonenl 
Thanks for the contribution, LGTM.
Could you please just sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw)?

laaksonenl:
Done

slandelle:
Thanks!

PR #3867, opened 2020-02-20
registerJmesPathFunctions always results in UnsupportedOperationExcep…

carstensandmadsen:
The check to see if `registerJmesPathFunctions` has already been called, incorrectly referred to a method parameter instead of an object variable and therefore did not allow registering new JmesPath functions even once.


slandelle:
cherry-picked, thanks!

PR #3878, opened 2020-03-11
Fix LogFileReader to not print to stdout

perlun:
I noted this while (ab)using the `LogFileReader` for a project of ours where I am converting the Gatling-generated `.log` file into a JSON format more suitable for subsequent processing. My initial idea was to redirect `stdout` from my console application to a `.json` file but these `println()` calls obviously made that harder.

The log level here can be discussed and I'm fine to change it to e.g. `logger.info()` calls if it's considered appropriate. My main concern is to get rid of the code that prints to stdout, since it's much easier for "machine-consumption" if we change these to be log statements instead.

Having that said, I _do_ realize that this might make things inconvenient for "standard" users of Gatling, where these stdout messages might be of use. To properly cater for both scenarios, one option would be to make `LogFileReader` take a stream as parameter, and print the messages to that stream instead. That way, it's still useful from both kinds of scenarios. If you'd like me to take this route, please let me know and give as much information as possible about the suggested implementation since I'm not particularly versed in Scala and "the Scala way" of doing things. 

Thanks in advance, and thanks again for a great load testing tool! :+1:

### DIFF
--- a/gatling-charts/src/main/scala/io/gatling/charts/stats/LogFileReader.scala
+++ b/gatling-charts/src/main/scala/io/gatling/charts/stats/LogFileReader.scala
@@ -47,7 +47,7 @@ class LogFileReader(runUuid: String)(implicit configuration: GatlingConfiguratio
 
   import LogFileReader._
 
-  println("Parsing log file(s)...")
+  logger.debug("Parsing log file(s)...")
 
   private val inputFiles = simulationLogDirectory(runUuid, create = false, configuration).files
     .collect { case file if file.filename.matches(SimulationFilesNamePattern) => file.path }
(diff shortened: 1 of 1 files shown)


slandelle:
Hi @perlun Thanks for reporting.
I went for a different solution where the println are moved to the app layer. I'm not ready to remove them as those (few) println are useful for all the users who launch manually, typically from the command line.

I don't know what you do with `LogFileReader` but beware it's an internal, not for public usage.

perlun:
> I went for a different solution where the println are moved to the app layer. I'm not ready to remove them as those (few) println are useful for all the users who launch manually, typically from the command line.

Totally fine with me. :+1: 

> I don't know what you do with LogFileReader but beware it's an internal, not for public usage.

:wink: Thanks for the warning. I wanted the `simulation.log` file in a more generically machine-readable format, so I used the existing Gatling `LogFileReader` to avoid having to write my own parser for the simulation log. My tool then emits a `simulation.json` file in the same folder. This file is then further analyzed by subsequent tools in our simulation project. (We want to run a number of Gatling tests, supplying different number of users/s etc, to find the set of simulation parameters that makes the number of "slow" requests increase above our configured threshold.)

I realize this could be a useful contribution to others as well. My tool is currently internal-only, but if you would be interested into incorporating it into Gatling proper, I could check if I would be allowed to release it as FOSS.

PR #3890, opened 2020-04-28
Recorder: Introduce config parameter 'recorder.http.useMethodAndUriAsPostfix'.

cgiorgetta:
### Motivation
When using Gatling recorder, the produced report is not that easy to read for us, as the request names look quite alike having name `request_$ID`.

For us it is helpful, having also request method and URI in the request name; the produced reports are now quite useful to us.
 
### Content of this PR
Introduce gatling-recorder config parameter `recorder.http.useMethodAndUriAsPostfix`. When set to true, use the HTTP method and the URI as a postfix for http(s) requests name.

E.g. request name will be `request_0:post_http://gatling.io/path1/file1` (instead of simply `request_0`). This improves legibility of recorded Gatling reports.

As this implementation sticked to approach of existing parameter `recorder.http.useSimulationAsPrefix`, it was quite a no-brainer.

### DIFF
--- a/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
+++ b/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
@@ -107,7 +107,8 @@ private[scenario] object RequestTemplate {
       else
         ""
     val prefix = if (config.http.useSimulationAsPrefix) simulationClass else "request"
-    s"""http("${prefix}_${request.id}")
+    val postfix = if (config.http.useMethodAndUriAsPostfix) s""":${request.method}_${request.uri}""" else ""
+    s"""http("${prefix}_${request.id}${postfix}")
 			.$renderMethod$renderHeaders$renderBodyOrParams$renderCredentials$renderResources$renderStatusCheck$renderResponseBodyCheck"""
   }
 
--- a/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
+++ b/gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
@@ -112,8 +112,8 @@ private[scenario] object RequestTemplate {
 			.$renderMethod$renderHeaders$renderBodyOrParams$renderCredentials$renderResources$renderStatusCheck$renderResponseBodyCheck"""
   }
 
-  def sanitizeRequestPostfix(name: String): String = {
-    s"""${name.replaceAll("[^-._=:/?&A-Za-z0-9]", "_")}"""
+  def sanitizeRequestPostfix(postfix: String): String = {
+    postfix.replaceAll("[^-._=:/?&A-Za-z0-9]", "_")
   }
 
   def render(simulationClass: String, request: RequestElement, extractedUri: ExtractedUris)(implicit config: RecorderConfiguration): String =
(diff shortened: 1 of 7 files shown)


sennl:
I like the additional configuration possibility. Will this PR be merged anytime soon?

slandelle:
Thanks for contributing.

I like the idea of adding some information regarding the request in the generated request name.
It could even be the only behaviour, instead of an option.

However, there are some shortcomings with your current implementation:
* the recorder generates also files for request bodies and optionally for response bodies. Those should be covered too and match the request name.
* we need some escaping so the request names don't end up breaking some JS in the HTML reports or causing invalid file names

luvarqpp:
Just suggestion, would it be hard to provide configuration, where user can provide [Gatling EL](https://gatling.io/docs/current/session/expression_el) string in configuration? This way, user can have "any" request name, depending on given simulation needs.

slandelle:
@luvarqpp Please don't highjack existing issues with unrelated topics.
High level suggestions and feature requests are best discussed on our [community mailing list](https://gatling.io/community-mailing-list/).

sennl:
I don't want to push to much, but I really like the functionality of this PR. @slandelle is it currently still not ok ?

gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
```diff
@@ -107,10 +107,15 @@ private[scenario] object RequestTemplate {
       else
         ""
     val prefix = if (config.http.useSimulationAsPrefix) simulationClass else "request"
-    s"""http("${prefix}_${request.id}")
+    val postfix = if (config.http.useMethodAndUriAsPostfix) s":${sanitizeRequestPostfix(s"${request.method}_${request.uri}")}" else ""
```
slandelle, line 110: `s":${sanitizeRequestPostfix(s"${request.method}_${request.uri}")}"` => `sanitizeRequestPostfix(s"${request.method}_${request.uri}"`

gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
```diff
@@ -107,10 +107,15 @@ private[scenario] object RequestTemplate {
...
+    s"""http("${prefix}_${request.id}${postfix}")
 			.$renderMethod$renderHeaders$renderBodyOrParams$renderCredentials$renderResources$renderStatusCheck$renderResponseBodyCheck"""
   }
 
+  def sanitizeRequestPostfix(postfix: String): String = {
+    postfix.replaceAll("[^-._=:/?&A-Za-z0-9]", "_")
+  }
```
slandelle, line 117: Please remove useless {}

slandelle:
@cgiorgetta  Sorry for the long delay. Could you please fix my last comments and sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw)?

gatling-recorder/src/main/scala/io/gatling/recorder/scenario/template/RequestTemplate.scala
cgiorgetta: You are right, this was too complicated. As I need `:` as prefix, I changed it to:

    ":" + sanitizeRequestPostfix(s"${request.method}_${request.uri}")

slandelle:
@cgiorgetta Thanks a lot for the updates! The last missing piece before merging is that you sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw).

slandelle:
cherry-picked, thanks a lot!

PR #3893, opened 2020-04-30
Charts: Remove background color for .head

perlun:
The previous way this was written causes the output to look quite ugly if the simulation results directory is quite long.

## Before the change (w/ Gatling 3.0.2)

As can be seen, the Gatling frontline infomercial overlays the results directory.

image

## After the change

(tested by flipping the CSS property as indicated)

image


slandelle:
cherry-picked, thanks!

PR #3938, opened 2020-08-24
add new assertion - deviatesAround, close #3847

### LINKED ISSUE #3847: Core: support relative percent deviation to "around" assertion target
I mean something like this:
```
.assertions(
    global.requestPerSec(target=10, deviation=.05, inclusive=true),

)
```
where "target" is target intensity and "deviation" is deviation in percent.

MaksSieve:
Add new assertion: deviatesAround for issue #3847.

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala
@@ -66,7 +66,13 @@ class AssertionWithPathAndTarget[T: Numeric](path: AssertionPath, target: Target
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   def around(mean: T, plusOrMinus: T, inclusive: Boolean = true): Assertion =
     between(numeric.minus(mean, plusOrMinus), numeric.plus(mean, plusOrMinus), inclusive)
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+  def deviatesAround(target: T, percentDeviationThreshold: Double, inclusive: Boolean = true): Assertion = {
+    val margin = numeric.fromInt((numeric.toDouble(target) * percentDeviationThreshold).floor.toInt)
+    between(numeric.minus(target, margin), numeric.plus(target, margin), inclusive)
+  }
   def is(value: T): Assertion = next(Is(numeric.toDouble(value)))
   def in(set: Set[T]): Assertion = next(In(set.map(numeric.toDouble).toList))
   def in(values: T*): Assertion = in(values.toSet)
+
 }
--- a/src/sphinx/cheat-sheet.yml
+++ b/src/sphinx/cheat-sheet.yml
@@ -567,9 +567,25 @@
         - keyword: between
           description: Check that the value of the statistic is between two thresholds
           syntax:
-            - signature: (thresholdMin,thresholdMax)
+            - signature: (thresholdMin, thresholdMax)
               description: the statistic's bounds
             - signature: (thresholdMin,thresholdMax, inclusive = false)
+              description: the statistic's bounds, excluded
+
+        - keyword: around
+          description: Check that the value of the metric is around a target value plus or minus a given margin.
+          syntax:
+            - signature: (value, margin)
+              description: the statistic's target and bounds margin
+            - signature: (value, margin, inclusive = false)
+              description: the statistic's bounds, excluded
+
+        - keyword: deviatesAround
+          description: Check that metric is around a target value plus or minus a given relative margin
+          syntax:
+            - signature: (value, relativeMargin)
+              description: the statistic's target and bounds margin
+            - signature: (thresholdMin, thresholdMax, inclusive = false)
               description: the statistic's bounds, excluded
 
         - keyword: is
@@ -583,7 +599,6 @@
           syntax:
             - signature: (sequence)
               description: the <i>sequence</i> the statistic has to be in
-
 
 - title: HTTP Action
   description: Define the HTTP requests sent in your scenario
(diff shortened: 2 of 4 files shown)


slandelle:
Thanks, but could you please add the missing compile test and documentation?

* https://github.com/gatling/gatling/blob/master/gatling-core/src/test/scala/io/gatling/core/compile/CoreCompileTest.scala
* https://github.com/gatling/gatling/blob/master/src/sphinx/general/assertions.rst
* https://github.com/gatling/gatling/blob/master/src/sphinx/cheat-sheet.yml

MaksSieve:
I've found that it needs to add some more changes to implement this feature. Exactly I'd like to change type parameter from `Int ` to `Double` for all methods here: https://github.com/gatling/gatling/blob/master/gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala#L36. The cause is that I need to threshold is need to be `Double` because it is percent not just multiplier.
Is this ok?

slandelle:
> Exactly I'd like to change type parameter from Int to Double for all methods here

No, that's not possible. That would be a break change that would cause issues in FrontLine.

gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala
```diff
@@ -66,7 +66,13 @@ class AssertionWithPathAndTarget[T: Numeric](path: AssertionPath, target: Target
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   def around(mean: T, plusOrMinus: T, inclusive: Boolean = true): Assertion =
     between(numeric.minus(mean, plusOrMinus), numeric.plus(mean, plusOrMinus), inclusive)
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+  def deviatesAround(target: T, percentDeviationThreshold: T, inclusive: Boolean = true): Assertion = {
```
slandelle, line 70: IMHO, percentDeviationThreshold should be a Double as it's a percentage.

slandelle (changes requested):

gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala
```diff
@@ -66,7 +66,13 @@ class AssertionWithPathAndTarget[T: Numeric](path: AssertionPath, target: Target
...
+    val margin = numeric.fromInt((numeric.toDouble(target) * numeric.toDouble(percentDeviationThreshold) / 100).toInt)
```
slandelle, line 71: Should we truncate or round?

gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala
MaksSieve: your right, it will be better

gatling-core/src/main/scala/io/gatling/core/assertion/AssertionBuilders.scala
MaksSieve: its a discussionable question, we can add another paramter, but i think it will make method overweight with parameters. changing general type to Int in future releases should remov this problem, but for now i prefer to use strickter borders with truncating

src/sphinx/cheat-sheet.yml
```diff
@@ -567,11 +567,27 @@
...
+              description: the statistic's target and bounds margin
+            - signature: (value, margin, inclusive = false)
+              description: the statistic's bounds, excluded
+
+        - keyword: deviatesAround
+          description: Check that metric is around a target value plus or minus a given relative margin
+          syntax:
+            - signature: (value, relativeMargin)
+              description: the statistic's target and bounds margin
+            - signature: (thresholdMin,thresholdMax, inclusive = false)
```
slandelle, line 588: nit: missing space

slandelle (changes requested):
One last nit and we're almost good.
The last thing is that we need you to please sign our Contributor Licence Agreement: https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw

MaksSieve:
Done 👍

src/sphinx/cheat-sheet.yml
MaksSieve: fixed!

slandelle:
@MaksSieve Please check CI: one of the files is not properly formatted :)

slandelle:
Running ;compile; test:compile would format your code automatically.

MaksSieve:
oh, sorry... now formated as required

slandelle (approved):

slandelle:
One last thing: could you please squash your commit into a single one and add a proper commit message with [this format](https://netty.io/wiki/writing-a-commit-message.html)?

src/sphinx/general/assertions.rst
```diff
@@ -115,6 +115,10 @@ Conditions can be chained to apply several conditions on the same metric.
 
 * ``around(value, plusOrMinus, inclusive = false)``: same as above but doesn't include bounds
 
+* ``deviatesAround(target, percentDeviationThreshold)``: check that metric is around a target value plus or minus a given relative margin
+
+* ``percentDeviationThreshold(target, percentDeviationThreshold, inclusive = false)``: same as above but doesn't include bounds
```
slandelle, line 120: deviatesAround, not percentDeviationThreshold

slandelle (changes requested):

src/sphinx/general/assertions.rst
MaksSieve: facepalm....

slandelle:
There were some weird things in the git history so I ended up squashing myself.
Thanks!

PR #3997, opened 2020-11-02
Core: Use event loop for pace

phiSgr:
Motivation:

With the virtual user <-> event loop affinity,
scheduling delays with the session event loop should be more performant.

Modifications:

`Pace`, like `Pause`, uses event loop for delay.
`SessionSpec.FakeEventLoop` uses a `java.util.Timer` to handle delays.

Result:

`Pace` does not depend on `ActorSystem`.

### DIFF
--- a/gatling-core/src/test/scala/io/gatling/core/session/SessionSpec.scala
+++ b/gatling-core/src/test/scala/io/gatling/core/session/SessionSpec.scala
@@ -17,7 +17,7 @@
 package io.gatling.core.session
 
 import java.{ util => ju }
-import java.util.{ concurrent => juc }
+import java.util.{ Timer, TimerTask, concurrent => juc }
 
 import io.gatling.BaseSpec
 import io.gatling.commons.stats.{ KO, OK }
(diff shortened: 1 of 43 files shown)


gatling-core/src/test/scala/io/gatling/core/session/SessionSpec.scala
```diff
@@ -28,10 +28,17 @@ import io.netty.channel.{ Channel, ChannelFuture, ChannelPromise, EventLoop, Eve
 import io.netty.util.concurrent.{ EventExecutor, Future => NFuture, ProgressivePromise, Promise, ScheduledFuture }
 
 object SessionSpec {
+  private val timer = new Timer
```
slandelle, line 31: We shouldn't have state in an object, even if it's for tests.
If FakeEventLoop becomes stateful and hold a resource such as a Timer, it can no longer be an object. It must be a class and we must create an instance in the tests.
The timer should be configured to a daemon Thread, not a user one.
The timer should be cancelled in a `afterAll` method.

slandelle (changes requested):
Thanks a lot!
Could you please have a look at my comments?

gatling-core/src/test/scala/io/gatling/core/session/SessionSpec.scala
phiSgr: The state inside the timer feels like the philosophy point of "a tree falls in a forest".

I forgot making the timer thread daemon. Didn't think much because the tests passed. 😅

slandelle: > The state inside the timer feels like the philosophy point of "a tree falls in a forest".

A Timer is a Thread, hence an important resource. I'd rather have it hidden inside a `FakeEventLoop` class and have tests own `FakeEventLoop` instances that they must cancel afterward.

The reason is that I think it would be nice to have FakeEventLoop used in other tests.

slandelle:
cherry-picked

Thanks a lot!

phiSgr:
With that done, I now feel comfortable asking for forgiveness for writing this perversion.

https://github.com/phiSgr/gatling-ext

PR #4021, opened 2020-12-03
Change the exit code to 1 on abnormal sys.exit() calls

### LINKED ISSUE #4005: Bundle: gatling.sh exits with 0 code when compilation fails
I run `gatling.sh` from a script. I encountered multiple occasions where Gatling failed, but the exit code was 0.

Expected behavior: exit code should be non-zero
Observed behavior: exit code 0

```
gatling.sh ....

        at scala.App.main(App.scala:80)
        at scala.App.main$(App.scala:78)
        at io.gatling.compiler.ZincCompiler$.main(ZincCompiler.scala:40)
        at io.gatling.compiler.ZincCompiler.main(ZincCompiler.scala)
There is no simulation script. Please check that your scripts are in user-files/simulations

$ echo $?
0
```

EugeneAbramchuk:
This will return exit code 1 in two cases:

- absence of simulations in user-files
- too many invalid attempts at choosing the simulation in interactive shell dialog

Closes #4005


slandelle:
Hi @EugeneAbramchuk 
Sorry, but the [commits](https://github.com/gatling/gatling/pull/4021/commits) in your PR are a mess.
Could you please update this PR (push force) or provide a new one with a single commit against [the single file you've changed](https://github.com/gatling/gatling/pull/4021/files), please?

EugeneAbramchuk:
Indeed, I merged gatling/master into my old fork and it all looked alright on the top of it.
I will make another one.

PR #4023, opened 2020-12-04
Return exit code 1 on premature system exit, close #4005

### LINKED ISSUE #4005: Bundle: gatling.sh exits with 0 code when compilation fails

EugeneAbramchuk:
This will return exit code 1 in two cases:

- absence of simulations in user-files
- too many invalid attempts at choosing the simulation in interactive shell dialog

Closes #4005

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/Selection.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/Selection.scala
@@ -101,7 +101,7 @@ object Selection {
       def readSimulationNumber(attempts: Int): Int = {
         if (attempts > MaxReadSimulationNumberAttempts) {
           println(s"Max attempts of reading simulation number ($MaxReadSimulationNumberAttempts) reached. Aborting.")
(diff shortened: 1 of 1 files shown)


EugeneAbramchuk:
Ok, I have recovered from a bout of git ineptitude (for now). For some odd reason, my master was 18 commits ahead of gatling/master.

slandelle:
Thanks!

PR #4036, opened 2021-01-05
Fix Brotli4j MacOS dependency

hyperxpro:
Dependency for `brotli4jMacOs` should be `native-macos_x86-64` instead of `native-linux_x86-64`.


slandelle:
Oops, thanks!

PR #4037, opened 2021-01-13
Websocket name based on Session contents

jklingelAvid:
Added support for websocket names based on Session contents
Added missing charset param in PebbleBenchmark.scala


slandelle:
Thanks a lot!

PR #4041, opened 2021-01-31
sbt 1.4.7

sullis:


slandelle:
Thanks, but no need to provide PRs for upgrading dependencies on this repository.
This repository is actually a downstream product of the private one where we push our work and scala-steward is configured there.

PR #4062, opened 2021-03-28
Use LongRef for ThreadLocal

phiSgr:
This is barely a micro-optimization (more like nano-optimization).
But minor cost like this can add up.

I have done a JMH test and confirmed a measurable speedup.


slandelle:
Oh, nice trick. I wasn't aware of this class, it's pretty convenient and I started using it.
Thanks a lot!

PR #4102, opened 2021-06-15
WebSockets: auto reply to specific messages before checks

pschiffe:
This is a follow up to #4100

To cover all cases when performing ws checks, we will process auto
replies at first, and only if there was no auto reply, we will process
ws checks.

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/action/ws/fsm/WsPerformingCheckState.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/action/ws/fsm/WsPerformingCheckState.scala
@@ -69,17 +69,19 @@ final case class WsPerformingCheckState(
   }
 
   override def onTextFrameReceived(message: String, timestamp: Long): NextWsState =
-    currentCheck match {
-      case WsTextFrameCheck(_, matchConditions, checks, _) =>
-        tryApplyingChecks(message, timestamp, matchConditions, checks)
+    if (autoReplyTextFrames(message, webSocket)) {
+      NextWsState(this)
+    } else {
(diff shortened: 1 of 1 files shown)


pschiffe:
Hi @slandelle , I'm not sure why, but it seems that the code here: https://github.com/gatling/gatling/blob/main/gatling-http/src/main/scala/io/gatling/http/action/ws/fsm/WsPerformingCheckState.scala#L76 seems to be ignored / not used during ws checks. I had to do also this modification to make it work properly.

slandelle:
Hi @pschiffe 

Sorry, I missed this during review.

However, your fix is too complicated.

The core issue is that `autoReplyTextFrames` should be the first thing that should be performed in `onTextFrameReceived`. If false, perform the check logic (you can remove `autoReplyTextFrames` from the match default branch).

Am I being clear?

pschiffe:
It depends on what should have higher priority. I was thinking that matching/checks should be more important than the autoreply, but maybe it doesn't really matter. I'll update the PR.

slandelle:
> but maybe it doesn't really matter

Yeah, I don't expect protocol heartbeat messages to be automatically replied to, and applicative messages that needs to be checked to overlap.

slandelle:
This won't compile, you must return a NextWsState.

```scala
if (autoReplyTextFrames(message, webSocket)) {
  NextWsState(this)
} else {
   ...
}
```

pschiffe:
Sorry, please take a look now, I've also tested it on our scenario.

slandelle:
Thanks a lot!

PR #4110, opened 2021-07-20
Override up to date check for all Gatling run tasks

scramados:
The documented way to override the up to date check for the gradle run task does not work for specific scenario commands. (i.e. gatlingRun-SimulationFQN)
By using the with type function this will also work for these commands.


Isammoc (approved):

slandelle:
Thanks a lot!

PR #4132, opened 2021-09-01
Update GatlingPropertiesBuilder.scala

gauravBINdas:
Change : import scala.collection.mutable._
Previously : import scala.collection.mutable

Fixed Issue : 
Exception in thread "main" java.lang.NoSuchMethodError: 'java.lang.Object scala.collection.mutable.Map$.empty()'
at io.gatling.core.config.GatlingPropertiesBuilder.<init>(GatlingPropertiesBuilder.scala:25)
at io.gatling.app.cli.ArgsParser.<init>(ArgsParser.scala:26)
at io.gatling.app.Gatling$.fromArgs(Gatling.scala:44)
at io.gatling.app.Gatling$.main(Gatling.scala:37)
at io.gatling.app.Gatling.main(Gatling.scala)

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/config/GatlingPropertiesBuilder.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/config/GatlingPropertiesBuilder.scala
@@ -16,7 +16,7 @@
 
 package io.gatling.core.config
 
-import scala.collection.mutable
+import scala.collection.mutable._
 
 import io.gatling.core.ConfigKeys._
 


slandelle:
This looks wrong to me. I suspect your issue is because you're not using the correct Scala version. Gatling currently requires Scala 2.13 and is not compatible with 2.12 or 3.

gauravBINdas:
@slandelle , Initially I was thinking the same but even after setting the scala version: 2.13.6 , I faced the same exception.
To resolve the exception , I think we need to import the Map trait that's why added   import scala.collection.mutable._

slandelle:
Could you please provide a way to reproduce your original issue?

slandelle:
Any update on this?

gauravBINdas:
The issue is fixed.Please discard the PR. You were right about the scala
version.gatling/lib folder was corrupted because of other scala version jar
files

Issue #4135, opened 2021-09-08
HTTP: support dynamic WebSocket check name


Amerousful:

toniisidoro:
+1

PR #4136, opened 2021-09-08
Add support EL for WS checkTextMessage/checkBinaryMessage

Amerousful:
https://github.com/gatling/gatling/issues/4135


slandelle (changes requested):
Sorry, but your approach is wrong. The Expression should be resolved in the Action, not in the FSM.

Amerousful:
Hmm ... Apparently I went the wrong way.
This should be done in WsSend, WsClose, etc., right?

slandelle:
Indeed. Meaning you would have to duplicate the WsFrameCheckSequenceBuilder, WsTextFrameCheck and WsBinaryFrameCheck classes :(

Could you please explain your use case?

Amerousful:
My use case is that I add a unique ID to the name of each request: build number + launch uniqueness. I add this through the EL. The need for this is that I then compare the metrics for different launches.

The problem is that checkTextMessage is written to logs / metrics and I cannot change its name through EL

Amerousful:
Stéphane, hello again!
I followed your advice and moved resolving to Actions.

But it seems to me that my decision is still "dirty"...
1) I don't know how to handle Failure correctly - https://github.com/gatling/gatling/pull/4136/files#diff-9b0ea77b65d56feb4d08135246b7be1851ca4d9fcd61a88d5a104246e3478572R43
2) Сonfuses with cast (asInstanceOf) - https://github.com/gatling/gatling/pull/4136/files#diff-9b0ea77b65d56feb4d08135246b7be1851ca4d9fcd61a88d5a104246e3478572R33

If you can give a hint or advice - I will be very grateful.

slandelle:
> My use case is that I add a unique ID to the name of each request: build number + launch uniqueness. I add this through the EL. The need for this is that I then compare the metrics for different launches.

With unique IDs in the check name, you're going to flood the reporting engine and make the reports unusable. I guess you expect the parse the log file yourself, which I definitely don't recommend as this file is an internal, subject to change anytime.

gatling-http-java/src/main/java/io/gatling/http/javaapi/Ws.java
```diff
@@ -153,23 +153,23 @@ private Prefix() {}
...
      * Bootstrap a check on inbound TEXT frames
      *
-     * @param name the name of the check
+     * @param name the name of the check, expressed as a Gatling Expression Language String
      * @return the next DSL step
      */
     @Nonnull
     public WsFrameCheck.Text checkTextMessage(@Nonnull String name) {
-      return new WsFrameCheck.Text(io.gatling.http.Predef.ws().checkTextMessage(name));
+      return new WsFrameCheck.Text(io.gatling.http.Predef.ws().checkTextMessage(toStringExpression(name)));
```
slandelle, line 161: Please also add the overload that takes a Java function.

gatling-http-java/src/main/java/io/gatling/http/javaapi/Ws.java
```diff
@@ -153,23 +153,23 @@ private Prefix() {}
...
      * Bootstrap a check on inbound BINARY frames
      *
-     * @param name the name of the check
+     * @param name the name of the check, expressed as a Gatling Expression Language String
      * @return the next DSL step
      */
     @Nonnull
     public WsFrameCheck.Binary checkBinaryMessage(@Nonnull String name) {
-      return new WsFrameCheck.Binary(io.gatling.http.Predef.ws().checkBinaryMessage(name));
+      return new WsFrameCheck.Binary(io.gatling.http.Predef.ws().checkBinaryMessage(toStringExpression(name)));
```
slandelle, line 172: Please also add the overload that takes a Java function.

gatling-http/src/main/scala/io/gatling/http/check/ws/WsFrameCheck.scala
```diff
@@ -30,7 +32,8 @@ sealed trait WsFrameCheck {
 }
 
 object WsFrameCheck {
-  final case class Binary(name: String, matchConditions: List[WsCheck.Binary], checks: List[WsCheck.Binary], isSilent: Boolean) extends WsFrameCheck {
+  final case class Binary(checkName: Expression[String], matchConditions: List[WsCheck.Binary], checks: List[WsCheck.Binary], isSilent: Boolean, name: String)
```
slandelle, line 35: I don't like it. I'd rather have 2 sets of types: one for the DSL with the Expression and one for runtime with the resolved value.

gatling-http/src/main/scala/io/gatling/http/check/ws/WsFrameCheck.scala
```diff
@@ -46,7 +49,8 @@ object WsFrameCheck {
       copy(isSilent = true)
   }
 
-  final case class Text(name: String, matchConditions: List[WsCheck.Text], checks: List[WsCheck.Text], isSilent: Boolean) extends WsFrameCheck {
+  final case class Text(checkName: Expression[String], matchConditions: List[WsCheck.Text], checks: List[WsCheck.Text], isSilent: Boolean, name: String)
+      extends WsFrameCheck {
```
slandelle, line 53: same as above

gatling-http/src/main/scala/io/gatling/http/action/ws/WsFrameCheckSequenceBuilder.scala
```diff
@@ -25,13 +25,29 @@ import io.gatling.http.check.ws.{ WsFrameCheck, WsFrameCheckSequence }
...
-        accValue <- acc
         timeout <- builder.timeout(session)
-      } yield accValue :+ WsFrameCheckSequence(timeout, builder.checks)
+        accValue <- acc
+      } yield accValue :+ WsFrameCheckSequence(timeout, resolveCheckName(builder.checks, session).asInstanceOf[List[T]])
     }
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.ListAppend"))
+  private def resolveCheckName[T <: WsFrameCheck](checks: List[WsFrameCheck], session: Session): List[WsFrameCheck] = {
```
slandelle, line 38: First, I don't like it. I'd rather have 2 sets of types: one for the DSL with the Expression and one for runtime with the resolved value.

Then, you don't properly handle failures. What you must do is a recursive resolution. See here for an example https://github.com/gatling/gatling/blob/15664aa2fbfbf4ac55190eb83115e219a777d529/gatling-core/src/main/scala/io/gatling/core/session/package.scala#L58-L79

slandelle (changes requested):

gatling-http/src/main/scala/io/gatling/http/action/ws/Ws.scala
```diff
@@ -31,9 +31,11 @@ object Ws {
 
   def apply(requestName: Expression[String], wsName: Expression[String]): Ws = new Ws(requestName, wsName)
 
-  def checkTextMessage(name: String): WsFrameCheck.Text = WsFrameCheck.Text(name, Nil, Nil, isSilent = false)
+  def checkTextMessage(name: Expression[String]): WsFrameCheck.Text = WsFrameCheck.Text(name, Nil, Nil, isSilent = false, name = "")
+  def checkTextMessage(name: String): WsFrameCheck.Text = checkTextMessage(itemToElString(name))
```
slandelle, line 35: Please remove. You don't need to overload: a String parameter would implicitly be converted into an Expression[String]. That's how **all** the other methods in Gatling's DSL work.

gatling-http/src/main/scala/io/gatling/http/action/ws/Ws.scala
```diff
@@ -31,9 +31,11 @@ object Ws {
...
 
-  def checkBinaryMessage(name: String): WsFrameCheck.Binary = WsFrameCheck.Binary(name, Nil, Nil, isSilent = false)
+  def checkBinaryMessage(name: Expression[String]): WsFrameCheck.Binary = WsFrameCheck.Binary(name, Nil, Nil, isSilent = false, name = "")
+  def checkBinaryMessage(name: String): WsFrameCheck.Binary = checkBinaryMessage(itemToElString(name))
```
slandelle, line 38: Please remove, same as above.

gatling-core/src/main/scala/io/gatling/core/session/package.scala
```diff
@@ -78,15 +78,22 @@ package object session {
...
   }
 
+  def itemToEl(item: Any): Session => Validation[Any] = {
+    item match {
+      case s: String => s.el[Any]
+      case v         => v.expressionSuccess
+    }
+  }
+
+  def itemToElString(item: String): Session => Validation[String] = {
```
slandelle, line 88: useless, see below, please revert

gatling-core/src/main/scala/io/gatling/core/session/package.scala
```diff
@@ -78,15 +78,22 @@ package object session {
     resolveRec(_, iterable.iterator, Nil)
   }
 
+  def itemToEl(item: Any): Session => Validation[Any] = {
```
slandelle, line 81: useless, see below, please revert

gatling-http/src/main/scala/io/gatling/http/action/ws/WsFrameCheckSequenceBuilder.scala
```diff
@@ -25,13 +26,43 @@ import io.gatling.http.check.ws.{ WsFrameCheck, WsFrameCheckSequence }
 object WsFrameCheckSequenceBuilder {
 
   @SuppressWarnings(Array("org.wartremover.warts.ListAppend"))
-  def resolve[T <: WsFrameCheck](builders: List[WsFrameCheckSequenceBuilder[T]], session: Session): Validation[List[WsFrameCheckSequence[T]]] =
+  def resolve[T <: WsFrameCheck](builders: List[WsFrameCheckSequenceBuilder[T]], session: Session): Validation[List[WsFrameCheckSequence[T]]] = {
```
slandelle, line 29: Please remove those useless extra curly braces.

gatling-http/src/main/scala/io/gatling/http/action/ws/WsFrameCheckSequenceBuilder.scala
```diff
@@ -25,13 +26,43 @@ import io.gatling.http.check.ws.{ WsFrameCheck, WsFrameCheckSequence }
...
+        checks <- resolveChecks(builder.checks, session)
+      } yield accValue :+ WsFrameCheckSequence(timeout, checks)
+    }
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.ListAppend"))
+  private def resolveChecks[T <: WsFrameCheck](checks: List[T], session: Session): Validation[List[T]] = {
+
+    @tailrec
+    def resolveCheckName[C <: WsFrameCheck](entries: Iterator[C], session: Session, acc: List[C]): Validation[List[C]] = {
```
slandelle, line 43: You mustn't use an Iterator for that, that's not the Scala way.
You must pattern match the List, see https://github.com/gatling/gatling/blob/main/gatling-core/src/main/scala/io/gatling/core/check/Check.scala#L47-L61 for an example

gatling-http/src/main/scala/io/gatling/http/check/ws/WsFrameCheck.scala
slandelle: > how similar logic was implemented for queryParamMap / queryParamSeq and decided to try it the same way. I found this to be a simpler solution...

Where is that?
`queryParam` uses 2 different sets:
* `io.gatling.http.request.builder.HttpParam` with unresolved Expressions
* `io.gatling.http.client.Param` for resolved Strings

See https://github.com/gatling/gatling/blob/main/gatling-http/src/main/scala/io/gatling/http/request/builder/package.scala#L62

slandelle (changes requested):

slandelle:
@Amerousful sorry I took so long to review. Do you think you'll be able to tackle my comments in the next days. We're drawing near the 3.7.0 release (probably the week after next week).

slandelle:
As the release is drawing near, I cherry-picked your commit, fixed rebasing and did a little clean up.
See https://github.com/gatling/gatling/commit/46eb42ec22e075f6a8e7b484f368a0ce79ce9b68

Thanks a lot for contributing!

Amerousful:
@slandelle 
Wow, that's cool!
Big thanks for your time to review the pull request. 
I will take time to check the clean-up in detail so I'll be able to write more beautiful code in the future.

PR #4156, opened 2021-10-26
doc(realtimemonitoring): Adding a process to connect gatling to influ…

yankoho:
The current documentation is working only for the previous version of Ifnluxdb 1. X. 

No need Grafana. With version 2 of influxdb, you need to use telegraf to load your data from graphite, and you can now create your dashboard in Influxdb directly.  If you need grafana, you can also connect grafana to influxdb 2.0 to use a flux connection.

### DIFF
--- a/src/docs/content/guides/realtime_monitoring/index.md
+++ b/src/docs/content/guides/realtime_monitoring/index.md
@@ -117,31 +117,9 @@ time                count max mean min percentiles50 percentiles75 percentiles95
 
 #### Install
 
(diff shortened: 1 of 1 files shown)


slandelle:
@yankoho Thanks for contributing.
Actually, I'm wondering if we shouldn't simply have pointer to the official InfluxDB documentation instead of maintaining our own tutorials for a third-party technology.
WDYT?

slandelle:
Sorry, but indeed, I went with removing any third party specific tutorial.
Thanks anyway.

PR #4173, opened 2021-12-01
Small refactoring for before/after methods

Amerousful:
Hey!
It's me again regarding 'before\after\ methods. I did some small improvements there. Probably, it will be useful or just a bit more readable? 
Will be great if you will be able to take a look at it?


slandelle:
Thanks! I've cherry-picked.

Issue #4220, opened 2022-02-26
Dos: The sample in the Quickstart is not working properly


Gezi-lzq:
When I compile the sample code(BasicSimlationJava) in the document, I can't run normally, an error occurs.
```
17:54:10.445 [ERROR] i.g.a.Gatling$ - Run crashed
java.lang.IllegalAccessException: class io.gatling.app.SimulationClass$Java cannot access a member of class gatlingtest.BasicSimlationJava with modifiers ""
```
I think there is a problem with setUp part brackets.
```
{
    setUp( // 11
      scn.injectOpen(atOnceUsers(1)) // 12
    ).protocols(httpProtocol); // 13
  }
```
I think it should be modified to：
```
    setUp( // 11
      scn.injectOpen(atOnceUsers(1)) // 12
         .protocols(httpProtocol)); // 13
  }
```
It can operate normally at this time.

slandelle:
I have no idea what you've done, but your analysis of your issue is wrong.

The `protocols` method is both available on the `SetUp` class and on the `PopulationBuilder` class so you can both define a global protocol and different ones per population.
Also, please note that our code samples are compiled so we wouldn't be able to release if there was a compile error there.

The error is something in your own code (your `gatlingtest.BasicSimlationJava` class). If you need help for figuring it out, please join our [community](https://community.gatling.io/) forum and post your code there.

PR #4221, opened 2022-02-26
Updata BasicSimlationJava.java

Gezi-lzq:
fix setUp
Issues: #4220


slandelle:
As explained here https://github.com/gatling/gatling/issues/4220#issuecomment-1052601184, the code sample is correct.

PR #4230, opened 2022-03-17
Doc: Fix links to Neo4j Bolt and JDBC plugins

geraldhumphries:
Minor documentation fix for broken links caused by rogue `>` characters.


slandelle:
Doh! Thanks!

PR #4244, opened 2022-05-18
fix: url in third_parties.md

Maksimall89:


slandelle:
Thanks, but this has already been fixed in main branch is this commit: https://github.com/gatling/gatling-private/commit/72895b79f4cebf641698d1ba44678e88f671c059 and the fix is already visible online: https://gatling.io/docs/gatling/reference/current/extensions/third_parties/.

Maksimall89:
Unfortunately, I can't see the commit from Gatling private. Also, I still see this bug.

изображение

slandelle:
> Unfortunately, I can't see the commit from Gatling private.

Sorry. Fixed link to public commit in my previous message.

> Also, I still see this bug.

Ha, you're correct. We'll publish the fix during the next release.

Maksimall89:
ok, thx

PR #4246, opened 2022-05-21
remove transform step if conversion to java is noop

phiSgr:
I was bothered seeing `.find.transform` in the logs when the user code has done no `transform`.

This change should also be a nano-optimization.


slandelle:
Hi @phiSgr 

Thanks a lot for reporting and contributing a fix proposal.

Actually, I wasn't fond of the identity comparison and went for a more in-depth fix, please see #4247.

Thanks again!

phiSgr:
I don't like the fragility of using object identity either. But not having wrapper function objects at runtime makes it worth it for me. Or maybe the JVM is smart enough to inline that, I don't know.

PR #4250, opened 2022-06-02
corrected typo (I think)

LloydWes:
Original sentence was "Supported formats are gzip and zip (but archive most contain only one single file)."
I think the word "must" was meant to be written instead of "most"


slandelle:
You're absolutely correct, thanks!
I've cherry-picked.

PR #4254, opened 2022-06-09
Update index.md

segaura:
tutorial is using Java 8 so the saved simulation file has "java" extension and not "scala"


slandelle:
Thanks a lot!
cherry-picked

PR #4314, opened 2022-08-12
Docs: Tutorial: Advanced: Kotlin example code should be more Kotlin-like

PaulienVa:
Make edit scenario  more Kotlin-like and use Kotlin functions and features instead of Java


slandelle:
Hi @PaulienVa ,

Please correct me if I'm wrong, but Kotlin's default random is just an alias for `java.util.Random`, meaning that it's slow because it's synchronized. That's the whole point of using `ThreadLocalRandom`.

PaulienVa:
Agh you are correct I overlooked that pretty crucial part 😋 , will discard the PR .

slandelle:
@PaulienVa There was also your second point about the useless  `session ->` function input parameter. I agree it could be removed, but we intentionally make it explicit to advertise for beginners that the input is the current virtual user's `Session` where they could extract stored data if needed.

PaulienVa:
Yes I figured that later on as well, the point was more about the random, in order to use a bit more of the Kotlin API ;)

Issue #4316, opened 2022-08-12
Core: add random numbers generators


slandelle:
Similar to #4315.

Original discussion here: #4305.

Let's first discuss the syntax here @Amerousful @shoaib42.

slandelle:
Original proposal from @Amerousful in https://github.com/gatling/gatling/pull/4311:

```
"#{randomNumber(10)}"
"#{randomNumber(5,10)}"
"#{randomNumber(5, max)}"
"#{randomNumber(5, maxL)}"
```

My comments on this:
1. all those would generate `Long`. Shouldn't we have distinct `randomInt` and `randomDouble`?
2. I'm not found of the magic `max` and `maxL`. I'd rather stick to explicit values.

I would suggest:

```
"#{randomInt()}" // full int range  
"#{randomInt(10)}" // between 0 and 10 inclusive
"#{randomInt(5, 10)}" // between 5 and 10 inclusive
"#{randomLong()}" // full long range   
"#{randomLong(10)}" // between 0 and 10 inclusive
"#{randomLong(5, 10)}" // between 5 and 10 inclusive
```

WDYT?

gemiusz:
I want to add something from my side:
I think that `randomDouble` will be beter but how to handle a precision if someone want to have 3.14 (not 3.1415...) - do you have thinking about using `BigDecimal` to handle all numbers ?
How about negative values ?
I don't thik that `full range` is a good idea I thik that `between X and Y` it's only good solution - this will be clear how it works when someone will see it.

shoaib42:
Late to join the party, here are my thoughts. I've listed them in the table below

| func/method  | use case | other requirements |
| ------------- | ------------- | ------------- |
| `randomLong()`  | full range -ve to +ve, counts/random ids ||
| `randomLong(min, max)`  | enumerated codes, "device type" etc, `randomLong(0, 10)` covers `randomLong(10)` ||
| `randomInt(min, max)`  |same as above||
| `randomDouble(min, max)`  | longitutde, latitude, we need negative floating points (OOTB is not covered), Device send slocation and response depends on this, might be a hot key that gets involved that needs stressing| needs +ve and -ve |
| `randomDouble(min, max, round)`  | currency, ending up in a DECIMAL in db| needs +ve and -ve |

I feel that adding too many flavours fo randomLong/randomDouble will turn into a long list for the endUser to know. That's my opinion, thought I throw that in.

[...]

PR #4317, opened 2022-08-15
Gatling EL: random Int and Long

Amerousful:
https://github.com/gatling/gatling/issues/4316

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
@@ -267,7 +267,10 @@ case object RandomInt extends ElPart[Int] {
 }
 
 final case class RandomIntRange(min: Int, max: Int) extends ElPart[Int] {
-  def apply(session: Session): Validation[Int] = (min + ThreadLocalRandom.current().nextInt((max - min) + 1)).success
+  require(min < max, s"Range `max` value must be above than `min`")
+  private val length = max - min + 1
+
+  def apply(session: Session): Validation[Int] = (min + ThreadLocalRandom.current().nextInt(length)).success
 }
 
 case object RandomLong extends ElPart[Long] {
@@ -275,8 +278,10 @@ case object RandomLong extends ElPart[Long] {
 }
 
 final case class RandomLongRange(min: Long, max: Long) extends ElPart[Long] {
+  require(min < max, s"Range `max` value must be above than `min`")
+  private val length = max - min + 1
 
-  def apply(session: Session): Validation[Long] = (min + ThreadLocalRandom.current().nextLong((max - min) + 1)).success
+  def apply(session: Session): Validation[Long] = (min + ThreadLocalRandom.current().nextLong(length)).success
 }
 
 class ElParserException(string: String, msg: String) extends Exception(s"Failed to parse $string with error '$msg'")
@@ -286,6 +291,7 @@ object ElCompiler extends StrictLogging {
   private val NameRegex = """[^.#{}()]+""".r
   private val DateFormatRegex = """[^#{}()]+""".r
   private val NumberRegex = """\d+""".r
+  private val NumberRegexWithNegative = """-?\d+""".r
   private val DynamicPartStart = "#{".toCharArray
 
   private val ElCompilers = new ThreadLocal[ElCompiler] {
@@ -414,13 +420,13 @@ final class ElCompiler private extends RegexParsers {
 
   private def randomInt: Parser[ElPart[Any]] = "randomInt()" ^^ (_ => RandomInt)
 
-  private def randomIntRange: Parser[ElPart[Any]] = "randomInt(" ~> NumberRegex ~ ("," ~> NumberRegex) <~ ")" ^^ { case min ~ max =>
+  private def randomIntRange: Parser[ElPart[Any]] = "randomInt(" ~> NumberRegexWithNegative ~ ("," ~> NumberRegexWithNegative) <~ ")" ^^ { case min ~ max =>
     RandomIntRange(min.toInt, max.toInt)
   }
 
   private def randomLong: Parser[ElPart[Any]] = "randomLong()" ^^ (_ => RandomLong)
 
-  private def randomLongRange: Parser[ElPart[Any]] = "randomLong(" ~> NumberRegex ~ ("," ~> NumberRegex) <~ ")" ^^ { case min ~ max =>
+  private def randomLongRange: Parser[ElPart[Any]] = "randomLong(" ~> NumberRegexWithNegative ~ ("," ~> NumberRegexWithNegative) <~ ")" ^^ { case min ~ max =>
     RandomLongRange(min.toLong, max.toLong)
   }
 
--- a/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
@@ -262,6 +262,42 @@ case object RandomUUID extends ElPart[UUID] {
   def apply(session: Session): Validation[UUID] = version4UUID().success
 }
 
+case object RandomInt extends ElPart[Int] {
+  def apply(session: Session): Validation[Int] = ThreadLocalRandom.current().nextInt().success
+}
+
+final case class RandomIntRange(arg: String) extends ElPart[Int] {
+
+  def randomInt: Int = {
+    val values = arg.split(",")
+    val min = values(0).toInt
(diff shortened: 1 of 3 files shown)


gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -395,7 +432,16 @@ final class ElCompiler private extends RegexParsers {
 
   private def randomUuid: Parser[ElPart[Any]] = "randomUuid()" ^^ (_ => RandomUUID)
 
-  private def nonSessionObject: Parser[ElPart[Any]] = currentTimeMillis | currentDate | randomUuid | randomSecureUuid
+  private def randomInt: Parser[ElPart[Any]] = "randomInt()" ^^ (_ => RandomInt)
+
+  private def randomIntRange: Parser[ElPart[Any]] = "randomInt(" ~> RangeRegex <~ ")" ^^ (n => RandomIntRange(n))
```
slandelle, line 437: Could you please try to perform the full parsing here instead of parsing the range in a second step?

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -262,6 +262,42 @@ case object RandomUUID extends ElPart[UUID] {
...
+case object RandomInt extends ElPart[Int] {
+  def apply(session: Session): Validation[Int] = ThreadLocalRandom.current().nextInt().success
+}
+
+final case class RandomIntRange(arg: String) extends ElPart[Int] {
+
+  def randomInt: Int = {
+    val values = arg.split(",")
+    val min = values(0).toInt
+    val max = values(1).toInt
```
slandelle, line 274: See my comment below for performing this parsing with parser combinators.
Then, 2 bad things here:
* the pattern doesn't allow for white spaces, eg `randomInt(5, 10)`, which users are very likely to write
* parsing should only be done once, not on each iteration

slandelle (changes requested):

slandelle:
@Amerousful Thanks. Could you please have a look at my comment?

Amerousful:
@slandelle In progress

slandelle:
>  I'm not sure about correctness of solution, but it works

Yes, this is the correct solution.

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -262,13 +262,30 @@ case object RandomUUID extends ElPart[UUID] {
   def apply(session: Session): Validation[UUID] = version4UUID().success
 }
 
+case object RandomInt extends ElPart[Int] {
+  def apply(session: Session): Validation[Int] = ThreadLocalRandom.current().nextInt().success
+}
+
+final case class RandomIntRange(min: Int, max: Int) extends ElPart[Int] {
+  def apply(session: Session): Validation[Int] = (min + ThreadLocalRandom.current().nextInt((max - min) + 1)).success
```
slandelle, line 270: Please:
* add some validation on the input parameters, that `min < max`, using `require`
* compute `max - min + 1` only once (private attribute, no need for the extra parens)

`val length = {
  require(max > min, "some error message")
  max - min + 1
}`

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -262,13 +262,30 @@ case object RandomUUID extends ElPart[UUID] {
...
+}
+
+case object RandomLong extends ElPart[Long] {
+  def apply(session: Session): Validation[Long] = ThreadLocalRandom.current().nextLong().success
+}
+
+final case class RandomLongRange(min: Long, max: Long) extends ElPart[Long] {
+
+  def apply(session: Session): Validation[Long] = (min + ThreadLocalRandom.current().nextLong((max - min) + 1)).success
```
slandelle, line 279: same as above

slandelle (changes requested):

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
Amerousful: @slandelle  Honestly, I thought about that, but I didn't find any other `require` there. I saw only some failure messages in `ElMessages` which related more to failures in EL logic, function, etc. And I decided maybe it's preferable to "trust" in exceptions which can be thrown by.

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -262,13 +262,30 @@ case object RandomUUID extends ElPart[UUID] {
...
+}
+
 class ElParserException(string: String, msg: String) extends Exception(s"Failed to parse $string with error '$msg'")
 
 object ElCompiler extends StrictLogging {
 
   private val NameRegex = """[^.#{}()]+""".r
   private val DateFormatRegex = """[^#{}()]+""".r
-  private val NumberRegex = "\\d+".r
+  private val NumberRegex = """\d+""".r
```
Amerousful, line 288: Need to add support negative numbers

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
slandelle: When possible, it's better to have some validation early when the String is parsed than having some Exceptions thrown when the Simulation is running.
If we spot other places where such requirements would make sense, let's add that in a follow-up PR.

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
slandelle: Indeed

slandelle: But then, please rename the existing regex into  PositiveNumberRegex to not break existing usage.

slandelle (changes requested):

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
Amerousful: Yep, I did it. But hadn't pushed yet

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -262,13 +262,36 @@ case object RandomUUID extends ElPart[UUID] {
...
+
+  def apply(session: Session): Validation[Int] = (min + ThreadLocalRandom.current().nextInt(length)).success
+}
+
+case object RandomLong extends ElPart[Long] {
+  def apply(session: Session): Validation[Long] = ThreadLocalRandom.current().nextLong().success
+}
+
+final case class RandomLongRange(min: Long, max: Long) extends ElPart[Long] {
+  require(min < max, s"Range `max` value must be above than `min`")
```
slandelle, line 281: Please add the min and max value in the error message.

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -262,13 +262,36 @@ case object RandomUUID extends ElPart[UUID] {
   def apply(session: Session): Validation[UUID] = version4UUID().success
 }
 
+case object RandomInt extends ElPart[Int] {
+  def apply(session: Session): Validation[Int] = ThreadLocalRandom.current().nextInt().success
+}
+
+final case class RandomIntRange(min: Int, max: Int) extends ElPart[Int] {
+  require(min < max, s"Range `max` value must be above than `min`")
```
slandelle, line 270: Please add the min and max value in the error message.

slandelle (changes requested):
Could you also please add the documentation of those new functions?

slandelle (changes requested):
Could you also please add the documentation of those new functions?

Amerousful:
Really, I forgot about documentation. Thank you.
Give me a minute)

Amerousful:
@slandelle hm, I have doubts about documentation. What is preferable? Just pattern or example with values?
1) `"#{randomLong(<min>,<max>)}"`
2) `"#{randomLong(2147483648,2147483658)}"`

slandelle:
@Amerousful With value with a comment like "where 2147483648 is the min and 2147483658 is the max" please

slandelle (approved):

slandelle:
Cherry-picked, thanks a lot!

PR #4318, opened 2022-08-17
fix sbt docs

phiSgr:
the command and the arg has to be in quotes, otherwise sbt sees it as two commands


guilgaly (approved):
LGTM, thanks a lot for the contribution!

slandelle:
Cheery-picked. Thanks a lot!

PR #4319, opened 2022-08-24
adds Nullable annotation to Session methods' return type

phiSgr:
Motivation:

Kotlin has nullable types, but it needs annotation from Java code to know if something is nullable.

Modifications:

Return types in the Session API are annotated @Nullable.
Comments in the Kotlin sample use the types as they are called in Kotlin.

Result:

Kotlin compiler now knows the nullability of values.

### DIFF
--- a/gatling-core-java/src/main/java/io/gatling/javaapi/core/Session.java
+++ b/gatling-core-java/src/main/java/io/gatling/javaapi/core/Session.java
@@ -44,8 +44,7 @@ public final class Session {
    * @param <T> the type of the desired value
    * @return the value if it exists, null otherwise
    */
-  @Nullable
-  public <T> T get(@Nonnull String key) {
+  public <T> @Nullable T get(@Nonnull String key) {
     return wrapped.attributes().getOrElse(key, () -> null);
   }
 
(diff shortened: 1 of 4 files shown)


phiSgr:
In Kotlin, there's nullable (`?` suffix), not-null (no suffix), and "platform type" (`!` suffix in IDE).

Without the `Nonnull` or `Nullable` annotations, a Java function's return type get resolved to the platform type.
They can be used as if they're not null, or can be safe-called (`?.methodCall`).

When a platform type value is written to a not-null field, `kotlin.jvm.internal.Intrinsics.checkNotNull` is inserted.
So the old code `val myPojo: MyPojo = session.get("key")` will throw if it's null.

gatling-core-java/src/main/java/io/gatling/javaapi/core/Session.java
```diff
@@ -43,6 +44,7 @@ public Session(io.gatling.core.session.Session wrapped) {
    * @param <T> the type of the desired value
    * @return the value if it exists, null otherwise
    */
+  @Nullable
```
slandelle, line 47: For consistency:
```java
public <T> @Nullable T get(@Nonnull String key)
```

slandelle (changes requested):
Thanks a lot!
Just one small nit.

gatling-core-java/src/main/java/io/gatling/javaapi/core/Session.java
phiSgr: Interesting. That was what I wrote in the first place, but IntelliJ warned "'Nullable' not applicable to type use".
It compiles fine apparently.

slandelle (approved):

slandelle:
Cherry-picked. Thanks!

PR #4320, opened 2022-08-24
add randomDouble with range and max fractional digits, close #4316

### LINKED ISSUE #4316: Core: add random numbers generators

shoaib42:
Feature : Add randomDouble generation provided as EL functions.

Functions added : 

1. `randomDouble(min, max)`, generates a random double in the range min (inclusive), max (exclusive)
2. `randomDouble(min, max, fractionalDigits)`, generates a random double in the range min (inclusive), max (exclusive) with at max  fractionalDigits

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
@@ -266,9 +266,7 @@ case object RandomInt extends ElPart[Int] {
 
 final case class RandomIntRange(min: Int, max: Int) extends ElPart[Int] {
   require(min < max, s"Range 'max'($max) must be above than 'min'($min)")
-  private val length = max - min + 1
-
-  def apply(session: Session): Validation[Int] = (min + ThreadLocalRandom.current().nextInt(length)).success
+  def apply(session: Session): Validation[Int] = ThreadLocalRandom.current().nextInt(min, max).success
 }
 
 case object RandomLong extends ElPart[Long] {
@@ -277,9 +275,22 @@ case object RandomLong extends ElPart[Long] {
 
 final case class RandomLongRange(min: Long, max: Long) extends ElPart[Long] {
   require(min < max, s"Range 'max'($max) must be above than 'min'($min)")
-  private val length = max - min + 1
+  def apply(session: Session): Validation[Long] = ThreadLocalRandom.current().nextLong(min, max).success
+}
 
-  def apply(session: Session): Validation[Long] = (min + ThreadLocalRandom.current().nextLong(length)).success
+final case class RandomDoubleRange(min: Double, max: Double) extends ElPart[Double] {
+  require(min < max, s"'max'($max) should be greater than 'min'($min)")
+  def apply(session: Session): Validation[Double] =
+    ThreadLocalRandom.current().nextDouble(min, max).success
+}
+
+final case class RandomDoubleRangeDigits(min: Double, max: Double, fractionalDigits: Int) extends ElPart[Double] {
+  require(min < max, s"'max'($max) should be greater than 'min'($min)")
+  require(fractionalDigits >= 0, s"fractionalDigits($fractionalDigits) must be positive")
+
+  private val tens = Math.pow(10.0, fractionalDigits)
+  def apply(session: Session): Validation[Double] =
+    ((ThreadLocalRandom.current().nextDouble(min, max) * tens).round / tens).success
 }
 
 class ElParserException(string: String, msg: String) extends Exception(s"Failed to parse $string with error '$msg'")
@@ -289,6 +300,7 @@ object ElCompiler extends StrictLogging {
   private val DateFormatRegex = """[^#{}()]+""".r
   private val NumberRegex = """\d+""".r
   private val NumberRegexWithNegative = """-?\d+""".r
+  private val DecimalRegexWithNegative = """-?\d+\.\d+""".r
   private val DynamicPartStart = "#{".toCharArray
 
   private val ElCompilers = new ThreadLocal[ElCompiler] {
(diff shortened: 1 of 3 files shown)


gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -284,6 +284,24 @@ final case class RandomLongRange(min: Long, max: Long) extends ElPart[Long] {
...
 
+final case class RandomDoubleRange(min: Double, max: Double) extends ElPart[Double] {
+  require(min < max, s"'max'($max) should be greater than 'min'($min)")
+  private val minInclusive = min
+  private val maxExclusive = max
+  def apply(session: Session): Validation[Double] =
+    ThreadLocalRandom.current().nextDouble(minInclusive, maxExclusive).success
+}
+
+final case class RandomDoubleRangeDigits(min: Double, max: Double, fractionalDigits: Int) extends ElPart[Double] {
```
slandelle, line 295: Wondering if the only use case for this is not just a random amount of currency (where fractionalDigits is always 2 and we don't deal with huge amounts that might overflow).

shoaib42: That's what I proposed for something of the sorts of currency or latitude/longitude, it can be time in seconds.nanoseconds as well. I don't think we will deal with large or small floating points. We can put a warning in the docs that near max Double or epsilon this will result in unexpected results.

slandelle: OK, you're right.

slandelle (changes requested):
Could you please ad a mention in the documentation about the possibility of overflow for large values and/or large scale?

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
shoaib42: Will do two changes:

1. Add details on docs, with warnings about floating points (large, epsilon)
2. Change the `digit` to `scale`

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -284,6 +284,24 @@ final case class RandomLongRange(min: Long, max: Long) extends ElPart[Long] {
...
+  require(min < max, s"'max'($max) should be greater than 'min'($min)")
+  require(fractionalDigits >= 0, s"fractionalDigits(${fractionalDigits}) must be positive")
+  private val minInclusive = min
+  private val maxExclusive = max
```
slandelle, line 299: What do you need those references for?

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
```diff
@@ -292,6 +310,7 @@ object ElCompiler extends StrictLogging {
   private val DateFormatRegex = """[^#{}()]+""".r
   private val NumberRegex = """\d+""".r
   private val NumberRegexWithNegative = """-?\d+""".r
+  private val FloatingRegexWithNegative = """[+-]?(\d*\.)?\d+""".r
```
slandelle, line 313: This regex is not correct:
* it matches incorrect values, eg `.1` or `0.1.1`
* why do you allow for `+` while NumberRegexWithNegative doesn't 

What about `-?\d+(\.\d+)?`?

slandelle (changes requested):

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
shoaib42: Sorry about that, these are left overs, will remove these.

gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
shoaib42: I tried a couple before I settled to this. Still made a boo boo when I pushed. Here is the list

1. `+.14` and `.14` are both valid  strings representing floating points, using `-?\d+(\.\d+)?` fails both on the `+` and the missing `0` before the decimal
2. Earlier I had the start and end anchors on the the regex, but that would fail with the parser, I've tried the start anchor `^` and this seems to work with the regex I've used - `^[+-]?(\d*\.)?\d+` (matches `1.1` from `1.1.1`)

As I see it `+145` should not fail for integers as this is a valid string representing an integer/long. While the `.1` is a valid floating point and `".1".toDouble` gives a valid `0.1: scala.Double`

I definitely need to add more test cases here :)

shoaib42: I see my confusion here :) with the `+` I'll try something soon

slandelle:
@shoaib42 Gentle ping. Do you think you'll be able to complete the requested changes, or should I pull and complete myself?

shoaib42:
Apologies @slandelle, please give me a day.

slandelle:
@shoaib42 We're going to have a release soon, probably next week. If possible, I'd like to merge this new feature and make it 3.9.0 and not 3.8.5. If you don't have time to complete, no problem, really, please just let me know and I can do it :)

Amerousful:
@slandelle My five cents) Also we were taking about random alphanum string. I'd like to implement it

slandelle:
> Also we were taking about random alphanum string. I'd like to implement it

Great, thanks! Feel free to send a separate PR. But let's ship the WebSocket PR first 😉

Amerousful:
Yep, I have almost done PR with WS)

shoaib42:
Midnight oil tonight and I'll send what I can.

shoaib42:
@slandelle pushed the changes, mostly had to update the doc side, the code was sitting there :/ + battling merge

slandelle:
@shoaib42 @Amerousful There's an inconsistency here with `randomInt` and `randomLong` because they currently include the right bound, on contrary to `Random.nextInt` and `Random.nextLong` (we do a `+1`). I fail to see a wait to include the right bound for `randomDouble`.

I feel like we should be consistent across random generators, meaning we should exclude the right bound everywhere (and have the same behavior as the underlying random generators). WDYT?

shoaib42:
@slandelle In JVM based languages I only know of random with range methods exclude the right bound and are always strictly `[min, max)` . (I do know that python `randint` includes the right bound).

The way to include the right bound on double would be to do something like `max + Double.MIN_VALUE`. 

What I think here is to keep it consistent with the underlying generators.

I can make these changes for the randomInt and randomLong - to exclude the right bound - in this PR

Amerousful:
Hmm, you're right... Perhaps, we should add words `to` and `until` to this EL functions as in `scala.collection.immutable.Range` and then the user will choose what preferred for his case
Or maybe I'm just overcomplicating things :)

Amerousful:
I've just remembered, we were taking about `inclusive` 
https://github.com/gatling/gatling/issues/4316#issuecomment-1213225067

slandelle:
> I've just remembered, we were taking about inclusive

Yeah, my bad. I didn't realize how the Java Random API was working and that we wouldn't be able to achieve it for random double. Let's favor consistency.

shoaib42:
Just clarifying, for consistency we will be changing randomInt and randomLong to make it right bound exclusive? I can make that change and add documentation for that.

Or is it that we are going to make randomDouble right bound inclusive? I find this less favourable as we are deviating from the Random API.

slandelle:
@shoaib42 the former please. Again, sorry for the original bad decision.

shoaib42:
All good, it's all good conversation. Please give me a few hours and I'll push (squashed).

slandelle:
@shoaib42 Also, we really must state explicitly in the documentation which bound is included and which is excluded, with the words.

The range notation differs from one country to another. 
* `[1, 5)` is the english notation
* in France, we would write `[1, 5[`, see https://fr.wikipedia.org/wiki/Intervalle_(math%C3%A9matiques)

I had to search the internet to figure out that the notation wasn't universal and that your `)` was not a mistake 🤯

shoaib42:
Who knew notations aren't universal! I agree, an improper expression and comprehension may lead from a "petit peu problème" into a  "pretty big problem".

slandelle (approved):

slandelle:
cherry-picked

Thanks a lot!

PR #4331, opened 2022-09-15
fix: ZonedDate for Gatling EL 'currentDate'

Amerousful:
Issue: https://github.com/gatling/gatling/issues/4332

```DateTimeFormatter.ofPattern(pattern).format(LocalDateTime.now())``` doesn't support time zone

Pattern ```EEE MMM dd YYYY HH:mm:ss 'GMT'Z``` will throw ```java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: OffsetSeconds```

### DIFF
--- a/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
+++ b/gatling-core/src/main/scala/io/gatling/core/session/el/ElCompiler.scala
@@ -17,7 +17,7 @@
 package io.gatling.core.session.el
 
 import java.{ util => ju }
-import java.time.LocalDateTime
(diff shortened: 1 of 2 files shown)


slandelle (changes requested):
Thanks a lot!
Could you also please fix the doc and squash your commits?
https://github.com/gatling/gatling/blob/main/src/docs/content/reference/current/core/session/el/index.md?plain=1#L79

Amerousful:
@slandelle You're welcome) Done!

slandelle:
cherry-picked, thanks!

PR #4335, opened 2022-09-16
fix: typo referencing Java property to debug TLS

### LINKED ISSUE #4334: Gatling.conf has typo for Java TLS debugging
The gatling.conf contains the text: 

> useOpenSsl = true                    # if OpenSSL should be used instead of JSSE (only the latter can be debugged with -Djava.net.debug=ssl)

Unfortunately this Java property is missing an "x", and therefore doesn't work as described. The correct property is `javax.net.debug`, see https://docs.oracle.com/javase/7/docs/technotes/guides/security/jsse/ReadDebug.html

As this took me a while to figure it out after I want to safe myself and other that time by fixing the typo. 

I'll prepare a PR for this.

ahus1:
Closes #4334

### DIFF
--- a/gatling-core/src/main/resources/gatling-defaults.conf
+++ b/gatling-core/src/main/resources/gatling-defaults.conf
@@ -50,7 +50,7 @@ gatling {
     maxThreadLocalCharBufferSize = 200000  # Netty's default is 16k
   }
   ssl {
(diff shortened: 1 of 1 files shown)


slandelle:
Cherry-picked, thanks a lot!

PR #4346, opened 2022-10-28
Update index.md

3alster:
Kotlin editor is not needed for Bundle, as Kotlin is not supported in a Bundle.  
Maven->Sbt fix copy-paste typo.


slandelle:
Thanks a lot!

PR #4351, opened 2022-11-17
Remove mnogu/gatling-kafka from Lagging

reftel:
It´s already (correctly) listed as Inactive.


slandelle:
cherry-picked, thanks!

PR #4355, opened 2022-11-23
Update Gatling Git status

barbasa:
The gatling-git [1] extension has been updated to use Galing 3.8, hence it can be moved back among the active ones. 

[1]: https://github.com/GerritForge/gatling-git


slandelle:
cherry-picked

Glad you managed to upgrade!

PR #4364, opened 2022-12-16
Websocket: fix date format (zoneId)

Amerousful:
I've caught the bug about zoneId for WS buffer. And I have no clue why I didn't catch it when run tests locally 

image 446


slandelle:
The `java.time` API is terrible. So many runtime failures that should be detected at compile time.

Thanks!

slandelle:
cherry-picked, thanks!

Amerousful:
> The `java.time` API is terrible. So many runtime failures that should be detected at compile time.
I found! I did't re-test after this changes https://github.com/gatling/gatling/pull/4307#discussion_r1035356583

Issue #4369, opened 2022-12-28
JMS: impossible to use built-in EL function `#{randomUuid()}` in JMS property


faisal6621:
When I am trying to use the [built-in function](https://gatling.io/docs/gatling/reference/current/core/session/el/#built-in-functions) `#{randomUuid()}` as:

```java
    scenario("jms")
        .exec(
            jms("message")
                .send()
                .queue("gatling")
                .textMessage("hello gatling")
                .property("random", "#{randomUuid()}")
        );
```

I am getting the following error:

```
javax.jms.MessageFormatException: Only objectified primitive objects, String, Map and List types are allowed but was: 9a2df475-05d7-0a3e-8e04-56351642b403 type: class java.util.UUID
```

I tried changing the property declaration to `.property("random", "#{randomUuid().toString()}")` but that does not work either.
```
io.gatling.core.session.el.ElParserException: Failed to parse #{randomUuid().toString()} with error ''}' expected but '.' found'
```

slandelle:
@Amerousful Maybe it was wrong to make `randomUuid()` return a `java.util. UUID` instead of a String. WDYT?

Isammoc:
@slandelle Shouldn't we have the `.toString()` builtin function?

slandelle:
@Isammoc I don't think so. 

1. Gatling EL is supposed to be a **simple** templating engine, not a full blown programming language where people can be under the impression they can insert any arbitrary Java code inside `#{}`. It feels like a dangerous rabbit hole.
2. @faisal6621 only tried this because he though things should "just work". I agree with this expectation. His `.toString()` was a workaround tentative.

In order to make things "just work", we have 2 possibilities:
1. make Gatling JMS support work for any property value type

Let's appreciate the type unsafety of `javax.jms.Message#setObjectProperty`

```java
    /** Sets a Java object property value with the specified name into the 
      * message.
      *  
      * <P>Note that this method works only for the objectified primitive
      * object types ({@code Integer}, {@code Double}, 
      * {@code Long} ...) and {@code String} objects.
      *  
      * @param name the name of the Java object property
      * @param value the Java object property value to set
      *  
      * @exception JMSException if the JMS provider fails to set the property
      *                          due to some internal error.
      * @exception IllegalArgumentException if the name is null or if the name is
      *                          an empty string.
      * @exception MessageFormatException if the object is invalid
      * @exception MessageNotWriteableException if properties are read-only
      */ 
    void
[...]

PR #4370, opened 2022-12-29
EL: randomUUID to String

Amerousful:


Isammoc:
@Amerousful What about documentation?

Amerousful:
@Isammoc Thx for this point. Can you check is it everything ok?

src/docs/content/reference/current/core/session/el/index.md
```diff
@@ -83,9 +83,11 @@ Gatling EL provide the following built-in functions:
 "#{foo.htmlUnescape()}"
 
 // generate a random UUID (fast but cryptographically insecure)
+// UUID will be provided as String
```
slandelle, line 86: `generate a random UUID String` shorter and clearer :)

src/docs/content/reference/current/core/session/el/index.md
```diff
@@ -83,9 +83,11 @@ Gatling EL provide the following built-in functions:
 "#{foo.htmlUnescape()}"
 
 // generate a random UUID (fast but cryptographically insecure)
+// UUID will be provided as String        
 "#{randomUuid()}"
 
 // generate a random UUID (standard java.util.UUID#randomUUID: cryptographically secure, but slower)
+// UUID will be provided as String
```
slandelle, line 90: same as above

slandelle (changes requested):
Please also improve commit message, eg:

```
Make `randomUuid()` and `randomSecureUuid()` Gatling EL functions generate Strings and not java.util.UUID, close [#4369](https://github.com/gatling/gatling-private/issues/4369)

Isammoc (commented):
1. This should requires only one commit
2. @slandelle Does it break binary compatibility (other than this obvious `randomUuid`)?

slandelle:
> This should requires only one commit

Yes, we will squash.

> @slandelle Does it break binary compatibility (other than this obvious randomUuid)?

No, it doesn't. ElCompiler is not directly exposed.
There's indeed a behavior change, but not binary incompatibility, that's why I aimed at 3.9.1 and not 3.10.0.

slandelle (approved):

slandelle:
cheery-picked and squashed.

Thanks a lot!

Issue #4371, opened 2022-12-30
Documentation: Realtime Monitoring - Link to not actual version of InfluxDB Telegraph’s Graphite input


gemiusz:
Acctual version of InfluxDB Telegraph is 1.24 but Documentation refer to 1.20:

https://github.com/gatling/gatling/blob/b4e431758e775a3887bf54d41b2758e9116b49d6/src/docs/content/guides/realtime_monitoring/index.md?plain=1#L59

I wondering if it will be better to set version to `latest`:
https://docs.influxdata.com/telegraf/latest/data_formats/input/graphite/

slandelle:
> I wondering if it will be better to set version to latest

Agree. Do you want to send a PR?

gemiusz:
I will try to prepare PR :)

gemiusz:
PR: https://github.com/gatling/gatling/pull/4372

PR #4372, opened 2022-12-30
doc: Update InfluxDB Telegraph Graphite input link

gemiusz:
As mentioned in Issue: https://github.com/gatling/gatling/issues/4371

Acctual version of InfluxDB Telegraph is 1.24 but Documentation refer to 1.20:

https://github.com/gatling/gatling/blob/b4e431758e775a3887bf54d41b2758e9116b49d6/src/docs/content/guides/realtime_monitoring/index.md?plain=1#L59

I wondering if it will be better to set version to `latest`:
https://docs.influxdata.com/telegraf/latest/data_formats/input/graphite/


slandelle:
cherry-picked

Thanks a lot!

PR #4380, opened 2023-01-25
feat: added svg to the ignore static resource list

arturbdr:
Could we, also, ignore the svg request as part of the static resource list?


slandelle:
Makes sense. Thanks!

slandelle:
cherry-picked

Thanks!

slandelle:
@arturbdr  Gatling 3.9.1 is out.

arturbdr:
Thank you @slandelle!

PR #4385, opened 2023-01-31
Use https protocol for demo site

lpmi-13:
The domain supports https, and there's no reason not to use it.


slandelle:
Yes there is: configuring the Recorder for https requires extra steps because of the certificate verification. This is something we will revisit once we have a browser plugin recorder.

lpmi-13:
Ah, got it. Thanks for clarifying!

PR #4390, opened 2023-02-15
spelling mistake in gatling-defaults.conf

s0h311:
it should be "timeout", not "timedout"


slandelle:
Thanks, but the actual fix would be "timed out" = "reply messages that have timed out".

But thanks for reporting. I've fixed the missing space.

PR #4401, opened 2023-03-15
Core: print the actual values in the case of failed assertions

pauljtremblay:
Hello gatling devs. I have noticed that the console output for a simulation's failed assertions are not including the actual value in `RunResultProcessor`. This is not in line with more formal assertion templates such as `AssertionsJUnitTemplate` or `AssertionsJsonTemplate`.

I request that in the case of a failed assertion the actual value gets included in output printed to stdout, as that output serves as a very useful "executive report" for the simulation. This change would be extra helpful in the case where the actual value is not found in the assert condition's range such as with the `around()`, `between()`, and `deviatesAround()` conditions.

Attaching a screenshot showing assertions produced by `RunResultProcessor` without the actual values present for context.

failed assertion not printing actual value

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
@@ -67,7 +67,10 @@ private final class RunResultProcessor(configuration: GatlingConfiguration) {
 
   private def runStatus(assertionResults: List[AssertionResult]): StatusCode = {
     val consolidatedAssertionResult = assertionResults.foldLeft(true) { (isValid, assertionResult) =>
-      println(s"${assertionResult.message} : ${assertionResult.result}")
+      if (assertionResult.result)
+        println(s"${assertionResult.message} : ${assertionResult.result}")
+      else
+        println(s"${assertionResult.message} : ${assertionResult.result} (actual : ${assertionResult.actualValue.getOrElse(-1.0)})")
       isValid && assertionResult.result
     }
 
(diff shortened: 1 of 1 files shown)


gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
```diff
@@ -67,7 +67,10 @@ private final class RunResultProcessor(configuration: GatlingConfiguration) {
 
   private def runStatus(assertionResults: List[AssertionResult]): StatusCode = {
     val consolidatedAssertionResult = assertionResults.foldLeft(true) { (isValid, assertionResult) =>
-      println(s"${assertionResult.message} : ${assertionResult.result}")
+      if (assertionResult.result)
+        println(s"${assertionResult.message} : ${assertionResult.result}")
+      else
+        println(s"${assertionResult.message} : ${assertionResult.result} (actual : ${assertionResult.actualValue.getOrElse(-1.0)})")
```
slandelle, line 73: IMO, `-1.0` is a magical value that doesn't convey any meaning.
If we don't have an actual value, I'd rather print `-` instead.

slandelle (changes requested):
Thanks for your contribution.
Could you please address my comment and sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw)?

gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
pauljtremblay: Replaced `-1.0` with `-` per feedback.

slandelle (approved):

slandelle:
cherry-picked, squashed and merged.

Thanks a lot!

PR #4410, opened 2023-04-28
Remove old EL using

Amerousful:
`${}` -> `#{}`

I was trying to find all places with old syntax, but maybe I didn't find everything...


slandelle:
cherry-picked

Thanks!

Issue #4411, opened 2023-04-28
HTTP: formParamSeq/formParamMap - EL resolving issue


Amerousful:
Version: 3.9.3

Methods: `formParamMap` and `formParamSeq`
These methods can't resolve **simultaneously** map/seq AND values inside. 

Here you are example when `myMap` will be resolved, and `someValue` not:
```
Session("someValue" -> "foo", "myMap" -> Map("key" -> "#{someValue}"))

formParamMap("#{myMap}")
```

slandelle:
Sorry, I way on my phone and didn't properly understand your issue.

Sadly, that's a path I don't want to follow.

I don't want to support recursive EL nor have EL works inside functions:
* this harms performance as EL parsing must then be performed at runtime on every resolution
* you want 1 level of recursivity, but why not 2, 3, 4, etc?
* we would have to support this across the whole codebase

Amerousful:
Oh, I see the point! Got it.

PR #4412, opened 2023-04-28
[Fix] resolve EL for formParamMap/formParamSeq

Amerousful:
https://github.com/gatling/gatling/issues/4411


slandelle:
Declined, sorry

Issue #4413, opened 2023-04-28
HTTP: crashing after passing Map() as EL into formParamMap(formParamSeq)


Amerousful:
Version: 3.9.3

Error:  `Exception while handling HTTP/1.1 crash, please report to Gatling maintainers`

STR:
```
Iterator.continually(Map("map" -> Map()))

.formParamMap("#{map}")
```

Stacktrace:
```
19:29:31.184 [DEBUG] i.g.h.e.r.DefaultStatsProcessor - Request 'get request' failed for user 1: j.l.StringIndexOutOfBoundsException: String index out of range: -1
19:29:31.191 [ERROR] i.g.h.c.i.HttpAppHandler - Exception while handling HTTP/1.1 crash, please report to Gatling maintainers
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.base/java.lang.AbstractStringBuilder.setLength(AbstractStringBuilder.java:275)
	at java.base/java.lang.StringBuilder.setLength(StringBuilder.java:85)
	at io.gatling.http.client.body.form.FormUrlEncodedRequestBody.encode(FormUrlEncodedRequestBody.java:64)
	at io.gatling.http.client.body.form.FormUrlEncodedRequestBody.toString(FormUrlEncodedRequestBody.java:114)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:168)
	at io.gatling.http.client.Request.toString(Request.java:196)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:168)
	at scala.collection.IterableOnceOps.addString(IterableOnce.scala:1184)
	at scala.collection.IterableOnceOps.addString$(IterableOnce.scala:1179)
	at scala.collection.AbstractIterator.addString(Iterator.scala:1293)
	at scala.collection.IterableOnceOps.mkString(IterableOnce.scala:1129)
	at scala.collection.IterableOnceOps.mkString$(IterableOnce.scala:1127)
	at scala.collection.AbstractIterator.mkString(Iterator.scala:1293)
	at scala.runtime.ScalaRunTime$._toString(ScalaRunTime.scala:156)
	at io.gatling.http.response.HttpFailure.toString(Response.scala:35)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:168)
[...]

PR #4414, opened 2023-05-01
[Fix] edge case with Map()/Seq() for formParamMap()/formParamSeq()

Amerousful:
https://github.com/gatling/gatling/issues/4413

### DIFF
--- a/gatling-http-client/src/main/java/io/gatling/http/client/body/form/FormUrlEncodedRequestBody.java
+++ b/gatling-http-client/src/main/java/io/gatling/http/client/body/form/FormUrlEncodedRequestBody.java
@@ -61,7 +61,11 @@ public final class FormUrlEncodedRequestBody extends RequestBody.Base<List<Param
     for (Param param : content) {
       encodeAndAppendFormParam(sb, param.getName(), param.getValue(), charset);
     }
-    sb.setLength(sb.length() - 1);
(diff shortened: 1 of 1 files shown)


gatling-http-client/src/main/java/io/gatling/http/client/body/form/FormUrlEncodedRequestBody.java
```diff
@@ -61,7 +61,11 @@ private StringBuilder encode() {
     for (Param param : content) {
       encodeAndAppendFormParam(sb, param.getName(), param.getValue(), charset);
     }
-    sb.setLength(sb.length() - 1);
+    if (sb.length() > 0) {
+      sb.setLength(sb.length() - 1);
+    } else {
+      sb.setLength(sb.length());
+    }
```
Isammoc, line 68:
```suggestion
    sb.setLength(Math.max(0, sb.length() - 1))
```

or

```suggestion
    if (sb.length() > 0) {
      sb.setLength(sb.length() - 1);
    }
```

no need to set to the current value, does it?

Amerousful: Oh, it makes sense. I checked and found that `StringBuilder` from the pool has a length - 0

PR #4424, opened 2023-05-22
reduce garbage in StatsHelper.buckets

phiSgr:
This change skips the `Vector` of boxed `Integer`s. IMO it's also a bit clearer.

Feel free to close if you think it's not worth merging.


slandelle:
Nice catch, thanks a lot

PR #4438, opened 2023-06-23
Use the system default timezone for Date in HTML report folders and Run information.

### LINKED ISSUE #4437: Please use the system default timezone for the report folder name and date in the run information.
I'm using Gatling 3.9.5.		
I am using Gatling in Japan and there is a 9 hour time difference.		
		
The console output and the horizontal axis of HTML report charts are displayed in Japan time.		
But below is fixed to GMT.		
	Name of the folder where the HTML report will be stored(SimulationName-yyyyMMddHHmmssSSS)	
	Date in Run Information on the Global tab of the HTML report(uuuu-MM-dd HH:mm:ss)	
		
I want it to be easier to find the HTML report I want to check from the time in the report folder, and I want to check the Run Information date in Japan time.		
So I want the above two to be assembled using the system default timezone.

kudoueiji:


slandelle:
Thanks.
cheery-picked on main branch

slandelle:
Note: We can't have this breaking change. We've realized that teams distributed over multiple timezones use UTC as a standard. The change introduced in the PR is going to be turned into an optional behavior.

PR #4456, opened 2023-09-01
Make link to report clickable in terminal

### LINKED ISSUE #4455: Reports: Link to file report is not clickable
After a scenario is executed Gatling will print:

```
Please open the following file: <path to file>
```

This is inconvenient as it requires copy/pasting the path into the system browser. If it printed the following instead it would be possible to open the report through the system browser by holding down `ctrl` and clicking the link:

```
Please open the following: file://<path to file>
```

stianst:
Closes #4455


slandelle:
Hi,

I'm afraid the problem is way more complicated than what you've tried (I've already tried what you've done and gave up for now).
The issue here is that there isn't A terminal.
What you've done works on YOUR terminal.
It doesn't work on others and worse, breaks what currently works: clickage on iTerm2 and can be copy-pasted into a browser.

We would need a matrix with the before and after result:
* on various Linux terminals, from zsh to the default one in the distro
* on MacOS: default and iTerm2
* on Windows
* in IntelliJ's embedded terminal

I don't know if [JLine](https://github.com/jline/jline3) could help with this.

PR #4461, opened 2023-09-09
doc: add to third party plugins list elasticsearch logger

Amerousful:
Hey there!
Do you mind if I add my logger, which parses Gatling's logs and sends them to Elasticsearch, to third-party plugins?
https://github.com/Amerousful/gatling-elasticsearch-logs


slandelle:
Hey,

Sorry, but we can't add this one.

First, the log file format is an internal to the HTML reports module. It has changed several times in the past and is likely to change again in the future (we're considering a binary format).

Then, new reporting features fall in the scope of Gatling Enterprise, so we can't advertise a competing solution.

Thanks for your understanding.

Amerousful:
I see. It's a pity, but ok)
thx anyway

PR #4478, opened 2023-11-22
Fractional user rate callout

stb13579:
Addresses the use case where requests per second are desired. Also small language improvements

### DIFF
--- a/src/docs/content/reference/current/core/injection/code/InjectionSampleKotlin.kt
+++ b/src/docs/content/reference/current/core/injection/code/InjectionSampleKotlin.kt
@@ -39,6 +39,15 @@ setUp(
 )
 //#open-injection
 
+//#fractional-user
+//Injects 1 user every 2 seconds for 60 seconds
+setUp(
+  scn.injectOpen(
+    constantUsersPerSec(0.5).during(60)
+  ).protocols(httpProtocol)
+);
+//#fractional-user
+    
 //#closed-injection
 setUp(
   scn.injectClosed(
--- a/src/docs/content/reference/current/core/injection/index.md
+++ b/src/docs/content/reference/current/core/injection/index.md
@@ -59,7 +59,8 @@ The building blocks for open model profile injection are:
 8. `stressPeakUsers(nbUsers).during(duration)`: Injects a given number of users following a smooth approximation of the [heaviside step function](http://en.wikipedia.org/wiki/Heaviside_step_function) stretched to a given duration.
 
 {{< alert tip >}}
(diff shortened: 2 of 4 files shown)


Isammoc (approved):

slandelle (changes requested):
Please also update the code samples to demonstrate fractional values.

stb13579:
> Please also update the code samples to demonstrate fractional values.

The code tabs were not working for me locally, but I think they are OK based on the other examples in the documentation.

src/docs/content/reference/current/core/injection/code/InjectionSampleKotlin.kt
```diff
@@ -39,6 +39,15 @@ setUp(
 )
 //#open-injection
 
+//#fractional-user
+//Injects 1 user every 2 seconds for 60 seconds
+setUp(
+  scn.injectOpen(
+    constantUsersPerSec(0.5).during(60)
+  ).protocols(httpProtocol)
+);
```
slandelle, line 48: no semi colon in Kotlin

src/docs/content/reference/current/core/injection/index.md
```diff
@@ -58,6 +58,11 @@ The building blocks for open model profile injection are:
 7. `rampUsersPerSec(rate1).to(rate2).during(duration).randomized`: Injects users from starting rate to target rate, defined in users per second, during a given duration. Users will be injected at randomized intervals.
 8. `stressPeakUsers(nbUsers).during(duration)`: Injects a given number of users following a smooth approximation of the [heaviside step function](http://en.wikipedia.org/wiki/Heaviside_step_function) stretched to a given duration.
 
+{{< alert tip >}}
+Fractional user injection rates are allowed. A common usecase is if you want to express the injection in terms of requests per second and not users per second.
```
slandelle, line 62: No found of this. It introduces a confusing between users per sec and requests per sec.
This feature is not directly related to rps.

src/docs/content/reference/current/core/injection/code/InjectionSampleKotlin.kt
slandelle: Do we want to add more code samples? Or just update the existing one with fractional values? https://github.com/gatling/gatling/blob/364da77b5702689a1cc685e50e1a5bf262ba84ae/src/docs/content/reference/current/core/injection/code/InjectionSampleJava.java#L37-L40

slandelle (changes requested):

PR #4480, opened 2023-11-23
3d parties: add another Kafka plugin

Amerousful:


slandelle:
cherry-picked!

Thanks for your hard work!

Amerousful:
Thank you too! I'm really glad that you appreciate my plugin! 

p.s. I'll continue with plugin, because I shoud add support for Protobuf, Avro, etc.

slandelle:
Quick question: why were you compelled to write a new plugin instead of using the one from Tinkoff?

Amerousful:
@slandelle 

1) In their plugin wasn't a proper structure or chain for sending messages, for example, send a message without a key. 
2) I have a concern regarding the `send` method which has too many arguments including `headers`. In my plugin, I have made sure that all message entities can be set independently and in a proper chain. A similar example is how we create an HTTP request where we add an URL and then we can add headers via methods, and so on.
There's a visualization. https://github.com/Amerousful/gatling-kafka#chain-for-build-a-request
3) Another thing, my plugin resolves consumer rebalancing. Link [(1)](https://github.com/Amerousful/gatling-kafka/blob/main/src/main/scala/io/github/amerousful/kafka/client/KafkaTrackerPoll.scala#L85) and [(2)](https://github.com/Amerousful/gatling-kafka/blob/main/src/main/scala/io/github/amerousful/kafka/client/WaitRebalancing.scala)
4) They also have a bug/issue with terminating a consumer(in TrackerPoll). The consumer doesn't stop after the end of a test, and you should kill a load/test process. [Link](https://github.com/Tinkoff/gatling-kafka-plugin/blob/master/src/main/scala/ru/tinkoff/gatling/kafka/client/TrackersPool.scala#L79)
5) I decided to add basic and most frequently used methods to configure producer/consumer configuration in a protocol. In this way, it's easier to set up your test without searching for the proper config key value. Just an example - [`credentials`](https://github.com/Amerousful/gatling-kafka/blob/main/src/main/scala/io/github/amerousful/kafka/protocol/KafkaProtocolBuilder.scala#L43C11-L43C11)

PR #4493, opened 2023-12-15
log actualValue regardless

mrdingma:
It would be helpful to show the actual value of the assertion regardless of whether it passed or failed.

### DIFF
--- a/gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
+++ b/gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
@@ -53,8 +53,6 @@ private final class RunResultProcessor(configuration: GatlingConfiguration) {
   private def reportsGenerationEnabled: Boolean =
     configuration.core.directory.reportsOnly.isDefined || (configuration.data.fileDataWriterEnabled && !configuration.reports.noReports)
 
-  private def alwaysPrintActualValueEnabled: Boolean = configuration.charting.printActualValue
-
   private def generateReports(runId: String, logFileData: LogFileData, assertionResults: List[AssertionResult]): Unit =
     if (reportsGenerationEnabled) {
       println("Generating reports...")
@@ -69,10 +67,8 @@ private final class RunResultProcessor(configuration: GatlingConfiguration) {
       val message = AssertionMessage.message(assertionResult.assertion)
       assertionResult match {
         case AssertionResult.Resolved(_, success, actualValue) =>
-          if (success && alwaysPrintActualValueEnabled) {
+          if (success) {
               println(s"$message : true (actual : $actualValue)")
-          } else if (success) {
-              println(s"$message : true")
           } else {
             println(s"$message : false (actual : $actualValue)")
           }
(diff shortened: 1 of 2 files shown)


mrdingma (commented):
agree

gatling-core/src/main/resources/gatling-defaults.conf
```diff
@@ -77,6 +77,7 @@ gatling {
     noReports = false       # When set to true, don't generate HTML reports
     maxPlotPerSeries = 1000 # Number of points per graph in Gatling reports
     useGroupDurationMetric = false  # Switch group timings from cumulated response time to group duration.
+    printActualValue = false   # When set to true, will print actual value of the assertion even if the assertion passes.
```
slandelle, line 80: No need for this option, let's print the actual value anyway

slandelle (changes requested):
Thanks.
Could you please remove the option and sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw/)?

Note: we'll start with the release process on Monday, so please hurry so this change doesn't miss it.

gatling-core/src/main/resources/gatling-defaults.conf
mrdingma: sounds good. Updated!

gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
```diff
@@ -68,7 +68,7 @@ private final class RunResultProcessor(configuration: GatlingConfiguration) {
       assertionResult match {
         case AssertionResult.Resolved(_, success, actualValue) =>
           if (success) {
-            println(s"$message : true")
+              println(s"$message : true (actual : $actualValue)")
```
slandelle, line 71: format error :)

slandelle (changes requested):

gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
mrdingma: ah ty. fixed

slandelle:
Cherry-picked and clean up, thanks!

gatling-app/src/main/scala/io/gatling/app/RunResultProcessor.scala
```diff
@@ -68,7 +68,7 @@ private final class RunResultProcessor(configuration: GatlingConfiguration) {
       assertionResult match {
         case AssertionResult.Resolved(_, success, actualValue) =>
           if (success) {
-            println(s"$message : true")
+            println(s"$message : true (actual : $actualValue)")
           } else {
             println(s"$message : false (actual : $actualValue)")
           }
```
Isammoc, line 74: Now, same thing as:
```suggestion
          println(s"$message : $success (actual : $actualValue)")
```

Pedantic note: no space before semicolon in english:
```suggestion
          println(s"$message: $success (actual: $actualValue)")
```

PR #4524, opened 2024-02-19
Change 'dandling' to 'dangling' in Session docs

marshblocker:


guilgaly:
Hi @marshblocker, thanks a lot for the contribution! We're currently doing a pretty big rework of the documentation, so I don't know if we'll merge your pull request as-is, but we'll make sure to include your fix either way.

PR #4526, opened 2024-02-22
Doc: Set correct markdown link to run-the-simulation-locally

gemiusz:


guilgaly:
Thanks, just cherry picked the fix! 👍

This one should have been caught when building the doc, because we have improved how `ref`s are validated - we'll look into why this could build without error.

PR #4531, opened 2024-02-26
doc: fix documentation url in README

BurzEX:
Original link refers to gatling.io/docs/current/ and redirecting to docs.gatling.io/current/ (it is 404 now). Actual docs location is docs.gatling.io


notdryft:
I've fixed the 404 but updated the link can't hurt. Cherry picked, thanks!

PR #4539, opened 2024-03-08
support for multiple business objects inside one jms message

artur-jablonski:
Hi. 

I came across a use case where application sends multiple "business-objects" in one JMS message and also gets batched response back. That is, a response can contain multiple business objects, but there is no guarantee that all business objects from request will be inside one response. Actually it's rather unlikely to happen. 

Each business object has unique id that is also send back in response. 

I tried to use Gatling for performance testing of this application, but it looks like the assumption is that one JMS message can only contain one id. 

I hacked around and modified the code and made it work, but this is breaking change as the JmsMessageMatcher interface changed as it's returning `Seq[String]` now instead of single `String`.

Can someone have a look and perhaps suggest how to add support for such feature in a backward compatible manner?

I was thinking perhaps to create new "batched" interface of matcher and provide some implicit conversion from the current Interface to the new "batched" interface. Any better ideas?


slandelle:
Closing. Can't be merged.

Issue #3368, opened 2017-10-20
Core: Investigate FastParse


slandelle:
We use Scala Parse Combinators for JsonPath and Gatling ElCompiler.

Compilation is usually not in the hot path as we cache results.
Still, we could investigate [FastParse](http://www.lihaoyi.com/fastparse/) and check if it's more convenient and faster.

http://www.lihaoyi.com/post/FastParse10PastPresentFuture.html

slandelle:
Sadly, I don't expect to have time for this any time soon, things work and parsing is usually not in the hot path (cached).

Closing for now.

PR #4560, opened 2024-05-08
Convert gatling-jsonpath to use FastParse

JD557:
Converts gatling-jsonpath to FastParse.

Related to #3368

My main motivation is not related to performance, however. The dependency on the parser-combinators 2.x, which makes `gatling-json` a bit annoying to use as a standalone library (as it's binary incompatible with 1.x, which clashes with some libraries that decided to not update until the next major, such as specs2).

As such, I didn't run any benchmarks. From my previous experience porting code from parser-combinators to FastParse I do expect this to be significantly faster.

I also think that the `ThreadLocal` in https://github.com/gatling/gatling/blob/376516cb4049904e1cf20712043f54b8087eebef/gatling-jsonpath/src/main/scala/io/gatling/jsonpath/JsonPath.scala#L33 is not required anymore (as `FastParse` keeps the mutable state in the implicit object being passed around).

I can change that as well, but I decided to keep it as I'm not sure if there are other reasons to use so many `ThreadLocal`s in the code.

### DIFF
--- a/gatling-jsonpath/src/main/scala/io/gatling/jsonpath/JsonPathParser.scala
+++ b/gatling-jsonpath/src/main/scala/io/gatling/jsonpath/JsonPathParser.scala
@@ -16,18 +16,23 @@
 
 package io.gatling.jsonpath
 
-import scala.util.parsing.combinator.RegexParsers
-
 import io.gatling.jsonpath.AST._
 import io.gatling.shared.util.StringBuilderPool
+
+import fastparse._
+import fastparse.MultiLineWhitespace._
 
 /**
  * Originally contributed by Nicolas Rémond.
  */
-private[jsonpath] object JsonPathParser extends RegexParsers {
+private[jsonpath] object JsonPathParser {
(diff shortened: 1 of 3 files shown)


slandelle:
Hi,

Thanks for the contribution.
Sadly, sorry, but for now I'm very reluctant to merge this.

I don't see any clear benefit for Gatling and only extra burden and risks.

1. We moved back this library into the Gatlong project and we wanted to make it clear that it was a Gatling internal and not intended to be used as a standalone library. If you manage to use it for your own needs, fine. But it's subject to changes without any prior notice.
2. Why change something that works?
3. Extra effort for us Gatling maintainers to learn Fastparse
4. Dependency to a libraries (fastparse and its dependencies) with one single maintainer (no offense meant to Li Haoyi)
5. Performance wise, in Gatling, JsonPath expression parsing rarely happens on the hot path. What happens on the hot path is the JSON parsing and applying the expression.

If you really need this, I think it would make more sense for you to write your own library (please don't forget to retain the copyright headers if you fork us).

Regards

PR #4561, opened 2024-05-10
Update InjectionSampleJava.java

ravigajul:
Updated randomized() to randomized on lines 38 and 40


ravigajul (commented):
updated randomized() to randomized on lines 38 and 40

slandelle:
This commit is wrong. Gatling Java requires these parens on contrary to Gatling Scala and Kotlin.

PR #4579, opened 2024-08-19
HTTP: Support recent "Secure Context" concept for secure cookies, thereby sending them also to localhost

### LINKED ISSUE #4578: HTTP: Gatling should send secure cookies to "localhost" as well
The spec: The W3C specs describe that secure context are those of https URLs, but also localhost and some others. 
https://w3c.github.io/webappsec-secure-contexts/#algorithms. Cookies marked as secure are sent to a secure context.

The expected behavior: When I run a Gatling test against `http://localhost:8000`, I expect Gatling to send secure cookies to the server.

Actual behavior: The current code only considers URLs starting with "https://" as a secure. 

https://github.com/gatling/gatling/blob/ad01135d2849a508d3a388a2dee9d405e4ac1ded/gatling-http-client/src/main/java/io/gatling/http/client/uri/Uri.java#L79

https://github.com/gatling/gatling/blob/610f0b6bcc7b226d98a1162dcc37b2f0d8ab1f44/gatling-http/src/main/scala/io/gatling/http/cookie/CookieJar.scala#L121-L127

In comarison, this is the logic Keycloak implemented recently around this to identify a secure context: https://github.com/keycloak/keycloak/blob/310824cc2b17eed2a48e93d7275efb4163cc9f6e/services/src/main/java/org/keycloak/cookie/SecureContextResolver.java

### With a browser like a Chrome 127 (both expected and observed behavior): 

* Prepare a server that sends Secure cookies: The included example sends a secure cookie to the browser in every request. Run with JDK 21 and a command `java CookieServer.java`.
* Open the URL in the browser. The first request does not send a in the  a cookie yet to it prints "Cookie Missing"
* Hit "refresh" to send another request to the sever. This second request contains the cookie, so it prints "Cookie Received".

### With Gatling: 

Expected behavior: First response is 400 / Cookie Missing, the second 200 / Cookie Received as it received the cookie

Actual behavior: The first and the second response are 400 / Cookie Missing (tested with Gatling 3.11.5 and the gatling-maven-plugin-demo-java-main project):

Prepare a server that sends Secure cookies: The included example sends a secure cookie to the browser in every request.
Run with JDK 21 and a command `java CookieServer.java`.

```
// CookieServer.java
public class CookieServer {

    public static void main(String[] args) throws Exception {
        HttpServer server = HttpServer.create(new InetSocketAddress(8000), 0);
        server.createContext("/", new Cookie());
        server.setExecutor(null);
        server.start();
        System.out.println("Server listening on port 8000");
    }

    static class Cookie implements HttpHandler {
        @Override
        public void handle(HttpExchange t) throws IOException {
            String response;
            int result;
            List<String> cookies = t.getRequestHeaders().get("Cookie");
            if (cookies != null && cookies.stream().anyMatch(s -> s.contains("mycookie=value"))) {
                response = "Cookie Received";
                result = 200;
            } else {
                response = "Cookie Missing";
                result = 400;
            }
            t.getResponseHeaders().put("Set-Cookie", List.of("mycookie=value; SameSite=None; Secure"));
            t.sendResponseHeaders(result, response.length());
            OutputStream os = t.getResponseBody();
            os.write(response.getBytes());
            os.close();
        }
    }
}

```

The scenario:

```
package computerdatabase;

import static io.gatling.javaapi.core.CoreDsl.*;
import static io.gatling.javaapi.http.HttpDsl.*;

import io.gatling.javaapi.core.*;
import io.gatling.javaapi.http.*;

/**
 * This sample is based on our official tutorials:
 * <ul>
 *   <li><a href="https://gatling.io/docs/gatling/tutorials/quickstart">Gatling quickstart tutorial</a>
 *   <li><a href="https://gatling.io/docs/gatling/tutorials/advanced">Gatling advanced tutorial</a>
 * </ul>
 */
public class CookieSimulation extends Simulation {

    ChainBuilder noCookieYet = exec(
        http("FirstRequestNoCookie").get("/")
                .check(status().is(400))
    );

    ChainBuilder withCookieSet = exec(
            http("SecondRequestWithCookie").get("/")
                    .check(status().is(200))
    );

    HttpProtocolBuilder httpProtocol =
        http.baseUrl("http://localhost:8000");

    ScenarioBuilder users = scenario("Users").exec(noCookieYet, withCookieSet);

    {
        setUp(
            users.injectOpen(rampUsers(1).during(10))
        ).protocols(httpProtocol);
    }
}

```

The output: 

```
> Global                                                   (OK=1      KO=1     )
> FirstRequestNoCookie                                     (OK=1      KO=0     )
> SecondRequestWithCookie                                  (OK=0      KO=1     )
```

ahus1:
Closes #4578

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/cookie/CookieJar.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/cookie/CookieJar.scala
@@ -118,12 +118,12 @@ private[http] final case class CookieJar(store: Map[CookieKey, StoredCookie]) {
     } else {
       val domain = requestDomain(requestUri)
       val path = requestUri.getNonEmptyPath
-      val secure = requestUri.isSecured
+      val secureContext = isSecureContext(requestUri)
 
       val matchingCookies = store.filter { case (key, storedCookie) =>
         domainsMatch(key.domain, domain, storedCookie.hostOnly) &&
         pathsMatch(key.path, path) &&
-        (secure || !storedCookie.cookie.isSecure)
+        (secureContext || !storedCookie.cookie.isSecure)
       }
 
       matchingCookies.toList
@@ -138,4 +138,27 @@ private[http] final case class CookieJar(store: Map[CookieKey, StoredCookie]) {
         }
         .map(_._2.cookie)
     }
+
+  private def isSecureContext(uri: Uri): Boolean =
+    if (uri.getScheme == Uri.HTTPS || uri.getScheme == Uri.WSS) {
+      true
+    } else {
+      val host: String = uri.getHost
+      if (host == null) {
+        false
+      } else if (host == "[::1]" || host == "[0000:0000:0000:0000:0000:0000:0000:0001]") {
+        // The host matches a CIDR notation of ::1/128// The host matches a CIDR notation of ::1/128
+        true
+      } else if (host.matches("127.\\d{1,3}.\\d{1,3}.\\d{1,3}")) {
+        // The host matches a CIDR notation of 127.0.0.0/8// The host matches a CIDR notation of 127.0.0.0/8
+        true
+      } else if (host == "localhost" || host == "localhost.") {
+        true
+      } else if (host.endsWith(".localhost") || host.endsWith(".localhost.")) {
+        true
+      } else {
+        false
+      }
+    }
+
 }
--- a/gatling-http/src/test/scala/io/gatling/http/cookie/CookieJarSpec.scala
+++ b/gatling-http/src/test/scala/io/gatling/http/cookie/CookieJarSpec.scala
@@ -284,6 +284,17 @@ class CookieJarSpec extends BaseSpec {
     cookies.head.isSecure shouldBe true
   }
 
+  it should "should send secure cookies to localhost" in {
+    // rfc6265#section-1 Cookies for a given host are shared  across all the ports on that host
+    val cookie1 = decode("cookie1=VALUE1; Path=/; Secure")
+    val cookieStore = CookieJar(Uri.create("http://localhost/moodle/"), List(cookie1), System.currentTimeMillis())
+
+    val cookies = cookieStore.get(Uri.create("http://localhost/moodle/login"))
+    cookies should have size 1
+    cookies.head.value shouldBe "VALUE1"
+    cookies.head.isSecure shouldBe true
+  }
+
   it should "should also serve non secure cookies based on the uri scheme" in {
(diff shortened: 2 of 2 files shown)


ahus1:
I'm not used to write Scala code, so it might need some rework from a Scala pro. At least the format check is now happy. 

I didn't change `requestUri.isSecured` as it is used in several other places that assume it check only https and then construct a URL with it, see: 

https://github.com/gatling/gatling/blob/24e0cec9b30810c06d68570116356cfe7766f694/gatling-http-client/src/main/java/io/gatling/http/client/util/HttpUtils.java#L71

Keeping the change local to the CookieJar was the approach of this PR. If you think the logic might be useful in other places of Gatling, a new property could be added to the `requestUri` instead.

Let me know your thoughs.

slandelle:
Thanks for your contribution. I’m currently off and will only be able to review next week.
Please stay tuned.
Thanks!

slandelle:
Thanks for the contribution. Could you please sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw/edit)?

I'll clean up/optimize the code afterward.

ahus1:
@slandelle - CLA signed. Thank you for cleaning up the code given my limited Scala skills.

gatling-http/src/main/scala/io/gatling/http/cookie/CookieJar.scala
```diff
@@ -138,4 +138,27 @@ private[http] final case class CookieJar(store: Map[CookieKey, StoredCookie]) {
...
+    if (uri.getScheme == Uri.HTTPS || uri.getScheme == Uri.WSS) {
+      true
+    } else {
+      val host: String = uri.getHost
+      if (host == null) {
+        false
+      } else if (host == "[::1]" || host == "[0000:0000:0000:0000:0000:0000:0000:0001]") {
+        // The host matches a CIDR notation of ::1/128// The host matches a CIDR notation of ::1/128
+        true
+      } else if (host.matches("127.\\d{1,3}.\\d{1,3}.\\d{1,3}")) {
```
slandelle, line 152: This line is not great performance-wise as you're compiling the underlying Pattern every time this method is executed. You might want to improve the implementation in Keycloak too.

slandelle:
I've cherry picked and cleaned up your changes. Thanks a lot!

gatling-http/src/main/scala/io/gatling/http/cookie/CookieJar.scala
ahus1: Thank you - this will be tracked in https://github.com/keycloak/keycloak/issues/32428

PR #4587, opened 2024-09-12
doc: fix documentation content

mir-jalal:
Gatling default configuration parameters are outdated in documentation


slandelle:
cherry-picked and published, thanks a lot!

Issue #4263, opened 2022-06-27
App: Invite users to upgrade Gatling version


slandelle:
Some users tend to not upgrade their Gatling version, possibly because they're not aware of the latest release and the new features and bug fixes.
Let's advertise the new releases in the console and add a deprecation warning in the HTML reports when the version is more than 1 year old.

clfsoft:
any options to turn this feature off ?
no one ever wanna a new version to replace working ones.

henricook:
@slandelle I'm thinking about raising a PR to disable this feature with a Java Opt (`-D`) - would you mind letting me know if this has already been discussed or you have any reason to think that the PR wouldn't be accepted?

Background: I have automatic dependency management and an allow-listed outbound environment and don't want to allow the communication required to check the release on every startup, or go through the seconds of wait time while it times out.

slandelle:
Sorry but this has already been discussed and we wouldn't accept this PR. We do want users to upgrade.

> go through the seconds of wait time while it times out.

The connect timeout and the read timeout are both 1s, so I expect the request to last at most 2s.

henricook:
@slandelle Darn sorry to hear that. It's a simple change with a default behaviour of `On` so it seems especially stern. It's hard to believe there are people out there without automatic dependency management taking care of this for them.

henricook:
On the topic of timeouts I should add that it failing due to timeouts on my local machine (UK based, 1Gbit up+down, generally low latency for a residential line) was what drew my eye to it in the first place.

slandelle:
> It's hard to believe (but I'm sure very real) that there are people out there without automatic dependency management taking care of this for them.

That's the vast majority.

> On the topic of timeouts I should add that it failing due to timeouts on my local machine

Weird. Do you have direct connection to the internet?

henricook:
I do indeed! Maybe it was a Maven burp as it's not deterministic and seems to mostly be working now 🤷🏻

PR #4588, opened 2024-09-24
Allow the gatling latest version check to be skipped

henricook:
This is a hygiene change driven by my own requirements. I'm in a locked down corporate environment and don't want Gatling hitting the outside on every startup. With automatic dependency management (e.g. Scala Steward) we also don't really need this check on startup.

Please let me know if I can update `docs.gatling.io` - I couldn't find a repository for it. And of course please let me know if I've missed any pull request etiquette and I'll gladly make changes.

Core of the change:

- Wrapped the version checking logic in an `if` based on a system prop:
```scala
if (java.lang.Boolean.getBoolean("disableLatestVersionCheck")) {...
```

Satisfies further comments in: https://github.com/gatling/gatling/issues/4263#issuecomment-1193397261

### DIFF
--- a/gatling-commons/src/main/scala/io/gatling/commons/util/GatlingVersion.scala
+++ b/gatling-commons/src/main/scala/io/gatling/commons/util/GatlingVersion.scala
@@ -146,23 +146,27 @@ private object LatestGatlingRelease extends StrictLogging {
       }
   }
 
-  def load(): Option[GatlingVersion] = {
-    val now = System.currentTimeMillis()
-    loadPersisted().logDebugOnFailure("Failed to load persisted latest release") match {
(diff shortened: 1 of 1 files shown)


slandelle:
Sorry, declining, as explained in response to your other message.

We want users to upgrade.
Sadly there are way more people not doing proper maintenance than people automating periodic upgrades, using excuses like "corporate rules" and "I don't have the budget nor the authority to upgrade".
This kind of behavior causes us extra burden, eg reaching out on our community forum about bugs that have been already fixed.
The option you're suggesting would exactly be used by this kind of users to stop being reminded they are not up-to-date: "stop bothering me and let me stick to outdated version".

henricook:
Thanks for the additional background

Issue #4443, opened 2023-07-19
Add support for jakarta.jms.ConnectionFactory


matzalew:
Hello Gatling team :)

We would like to be able to use jakarta.jms.ConnectionFactory as the JMS, as it's the valid library for newer versions of Springboot.
Currently it is not available.

slandelle:
Are you interested in contributing or sponsoring this feature (note: we can’t drop javax support either)?

slandelle:
Also, I'm a bit puzzled here. Unless I'm mistaken, the actual `ConnectionFactory` class you need is the one supported by your actual JMS client library (IBM MQS and friends). Springboot might be providing adapters, but still, unless you're using an updated client, you still have the old `javax` classes in your classpath.
And I really wonder how many JMS users have upgraded their JMS broker to move to the new "Jakarta EE" stack.

We'll most likely do this move someday, but when there will be enough demand for it.
As of today, you're the only OSS user to request it, and none of our Enterprise customers did.

slandelle:
@matzalew any update?

matzalew:
Sorry, missed your messages. 

Did not found workaround. 

I could try to contribute. But never done such thing, and I got likely limited time after my working hours. So no promises - but I'd like at least to try.

holomekc:
Hmm not sure how to tackle this. The only idea I have is to basically duplicate everything, and change the import when necessary, and use a different package:

![image](https://github.com/gatling/gatling/assets/30546982/3689d9a9-b836-4a56-ad48-0d83dbb4792f)

holomekc:
Normally other libraries create two releases. One with javax and one with jakarta. This makes the duplication unnecessary. e.g. 
gatling:3.10.4
gatling:3.10.4-jakarta

although I would like the other way around more:
gatling:3.10.4-javax
gatling:3.10.4

BusyByte:
@slandelle  any progress on this?   I would think since they are different artifacts you just have different artifact names.  you can add the one you want as a dependency.

slandelle:
@BusyByte Not a priority for us atm because:

* no customer demand, all our Enterprise customers who have a demand for JMS are still using a `javax` version.
* after only a quick glance, I couldn't find a way that would not boil down to duplicating the whole code, which I'm very reluctant to do

Such a stupid move from Oracle lawyers 😡

BusyByte:
[...]

PR #4603, opened 2025-01-03
add support for jakarta jms 3.1

BusyByte:
closes: https://github.com/gatling/gatling/issues/4443


slandelle:
[As I said](https://github.com/gatling/gatling/issues/4443#issuecomment-2569474740), this solution of forking everything should really be a last resort. Forking is easy, maintaining duplicated code is a huge pain. I think it should be possible to share the internals using some type classes to do the adapters, possibly inlined.

BusyByte:
@slandelle I was going to create another module named like jms common or core and move common stuff there and have the individual ones pull it in as a dependency

slandelle:
@BusyByte Great! Good luck!

PR #4606, opened 2025-01-21
chore: Enable gatling-http integration tests

j-madden:
Noticed that the `HttpIntegrationSpec` tests where ignored. When enable, they failed due to the following;
```
Cannot invoke "io.netty.channel.EventLoopGroup.next()" because "this.eventLoopGroup" is null
java.lang.NullPointerException: Cannot invoke "io.netty.channel.EventLoopGroup.next()" because "this.eventLoopGroup" is null
```
Resolved with using the `NioEventLoopGroup`.

The third test [ignore should "fetch resources in conditional comments" ](https://github.com/gatling/gatling/blob/9e58d329cccb9ee0745e988af5677345a2cd8eeb/gatling-http/src/test/scala/io/gatling/http/integration/HttpIntegrationSpec.scala#L127) is still failing as the conditional in the HTML is not triggering, kept it ignored.


gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
```diff
@@ -40,13 +40,15 @@ import io.netty.channel._
 import io.netty.handler.codec.http._
 import io.netty.handler.codec.http.cookie._
 import org.scalatest.BeforeAndAfter
+import org.scalatestplus.mockito.MockitoSugar
```
slandelle, line 43: Please rebase against current heap. You can no use NoopStatsEngine. We're trying to reduce Mockito usage, there's a chance we'll try to drop it some day.

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
```diff
@@ -40,13 +40,15 @@ import io.netty.channel._
...
 
 abstract class HttpSpec extends ActorSpec with BeforeAndAfter {
   type ChannelProcessor = ChannelHandlerContext => Unit
   type Handler = PartialFunction[FullHttpRequest, ChannelProcessor]
 
   private val clock = new DefaultClock
   protected val mockHttpPort: Int = Using(new ServerSocket(0))(_.getLocalPort).getOrElse(8072)
+  private val mockStatsEngine = MockitoSugar.mock[StatsEngine]
```
slandelle, line 51: see above

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
```diff
@@ -63,17 +65,18 @@ abstract class HttpSpec extends ActorSpec with BeforeAndAfter {
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   def runScenario(
       sb: ScenarioBuilder,
+      defaultSession: Session = emptySession,
```
slandelle, line 68: please remove, see below

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
```diff
@@ -63,17 +65,18 @@ abstract class HttpSpec extends ActorSpec with BeforeAndAfter {
...
     val protocols = Protocol.indexByType(Seq(protocolCustomizer(httpProtocol)))
     val coreComponents =
-      new CoreComponents(actorSystem, null, null, None, null, clock, null, configuration)
+      new CoreComponents(actorSystem, defaultSession.eventLoop, null, None, mockStatsEngine, clock, null, configuration)
     val protocolComponentsRegistry = new ProtocolComponentsRegistries(coreComponents, protocols).scenarioRegistry(Map.empty)
     val nextActor = mockActorRef[Session]("next")
     val next = new ActorDelegatingAction("next", nextActor)
     val action = sb.build(new ScenarioContext(coreComponents, protocolComponentsRegistry, Constant, throttled = false), next)
-    action ! emptySession
+    action ! defaultSession
```
slandelle, line 79: Please move in this class the EventLoopGroup you've added in HttpIntegrationSpec.
This way, you can do:
```
emptySession.copy(eventLoop = eventLoopGroup.next())
      .set(DnsCacheSupport.DnsNameResolverAttributeName, InetAddressNameResolver.JAVA_RESOLVER)
```

slandelle (changes requested):
Thanks a lot!

Please also remove the test about conditional block, this is something we've dropped.

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
j-madden: Sure thing, though implemented it as follows:
```
private val nioEventLoop: EventLoop = new NioEventLoopGroup().next()
...
emptySession.copy(eventLoop = nioEventLoop).set(DnsCacheSupport...)
```
As the `CoreComponents` requires the same eventLoop.

slandelle: Why not store in a val the EventLoopGroup instead of an EventLoop?

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
```diff
@@ -47,6 +50,7 @@ abstract class HttpSpec extends ActorSpec with BeforeAndAfter with EmptySession
 
   private val clock = new DefaultClock
   protected val mockHttpPort: Int = Using(new ServerSocket(0))(_.getLocalPort).getOrElse(8072)
+  private val nioEventLoop: EventLoop = new NioEventLoopGroup().next()
```
slandelle, line 53: private val nioEventLoopGroup = new NioEventLoopGroup()

slandelle (changes requested):

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
j-madden: Sorry, that's much cleaner. Will do.

slandelle (approved):

slandelle (changes requested):
Back to square one: these tests fail on CI while they work locally, see https://github.com/gatling/gatling/commit/e7a9657f8d86b3f59766eb1e56985cf2fbc697d8

No idea why they fail atm.

j-madden:
Really odd, works fine locally with sbt `1.10.7` & Azul `21.0.6` & the same environment variables.
It might be worth- increasing the log level to see why exactly it's failing on the CI.

gatling-http/src/test/scala/io/gatling/http/HttpSpec.scala
```diff
@@ -47,6 +50,7 @@ abstract class HttpSpec extends ActorSpec with BeforeAndAfter with EmptySession
 
   private val clock = new DefaultClock
   protected val mockHttpPort: Int = Using(new ServerSocket(0))(_.getLocalPort).getOrElse(8072)
+  private val nioEventLoopGroup: NioEventLoopGroup = new NioEventLoopGroup()
```
slandelle, line 53: Please replace with `private val eventLoopGroup = Transports.newEventLoopGroup(true, false, 1, "test")`.
On Linux (CI), the default channel type is epoll, not NIO, so it's not compatible with NioEventLoopGroup.

slandelle (changes requested):

slandelle (approved):

slandelle:
Cherry-picked. Thanks a lot!

PR #4640, opened 2025-09-16
feat: enable dynamic http method, close #4638

### LINKED ISSUE #4638: Support dynamic method type (ie. Expression[String]) in httpRequest signature
I know this can be done with `doSwitch` syntax, but only in the main request builder, not in request resources.

My use case is dynamically parsed HAR file (with variable requests and resources attached) that I want to replay in Gatling scenario. 

I think there will be more valid use cases for this as well :)

Example usage:

```
  val scenario = scenario("Scenario")
    .exec(_.set("main_http_method", "POST").set("resource_http_method", "GET"))
    .exec(
      http("Main")
        .httpRequest("#{main_http_method}", "/session")
        .resources(
          http("Resource")
            .httpRequest("#{resource_http_method}", "/session")
        )
    )

```

aartajew:
Motivation:

Be able to set HTTP request on the fly when replaying HAR file recorded requests. Another use case was described [here](https://stackoverflow.com/questions/37013490/gatling-httprequest-with-dynamic-method-from-collection).

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
@@ -34,7 +34,7 @@ final class Http(requestName: Expression[String]) {
   def head(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.HEAD, url)
   def delete(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.DELETE, url)
   def options(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.OPTIONS, url)
-  def httpRequest(method: Expression[String], url: Expression[String]): HttpRequestBuilder = httpRequest(Left(method), Left(url))
+  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(HttpMethod.valueOf(method)), Left(url))
   def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(method), Left(url))
   def httpRequest(method: Either[Expression[String], HttpMethod], urlOrURI: Either[Expression[String], Uri]): HttpRequestBuilder =
     HttpRequestBuilder(requestName, method, urlOrURI)
--- a/gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
@@ -27,15 +27,16 @@ import io.netty.handler.codec.http.HttpMethod
  */
 final class Http(requestName: Expression[String]) {
   def get(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.GET, url)
-  def get(uri: Uri): HttpRequestBuilder = httpRequest(HttpMethod.GET, Right(uri))
+  def get(uri: Uri): HttpRequestBuilder = httpRequest(Right(HttpMethod.GET), Right(uri))
   def put(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.PUT, url)
   def post(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.POST, url)
   def patch(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.PATCH, url)
   def head(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.HEAD, url)
   def delete(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.DELETE, url)
   def options(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.OPTIONS, url)
-  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.valueOf(method), Left(url))
-  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(method, Left(url))
-  def httpRequest(method: HttpMethod, urlOrURI: Either[Expression[String], Uri]): HttpRequestBuilder =
+  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(HttpMethod.valueOf(method)), Left(url))
+  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(method), Left(url))
+  def dynamicHttpRequest(method: Expression[String], url: Expression[String]): HttpRequestBuilder = httpRequest(Left(method), Left(url))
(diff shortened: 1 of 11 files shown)


gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
```diff
@@ -27,15 +27,16 @@ import io.netty.handler.codec.http.HttpMethod
...
   def patch(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.PATCH, url)
   def head(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.HEAD, url)
   def delete(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.DELETE, url)
   def options(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.OPTIONS, url)
-  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.valueOf(method), Left(url))
-  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(method, Left(url))
-  def httpRequest(method: HttpMethod, urlOrURI: Either[Expression[String], Uri]): HttpRequestBuilder =
+  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(HttpMethod.valueOf(method)), Left(url))
+  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(method), Left(url))
+  def dynamicHttpRequest(method: Expression[String], url: Expression[String]): HttpRequestBuilder = httpRequest(Left(method), Left(url))
```
slandelle, line 39: Sorry but we're trying to keep of public API tidy, so adding an extra method name doesn't work.

aartajew: Done

gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
```diff
@@ -27,15 +27,15 @@ import io.netty.handler.codec.http.HttpMethod
...
   def put(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.PUT, url)
   def post(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.POST, url)
   def patch(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.PATCH, url)
   def head(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.HEAD, url)
   def delete(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.DELETE, url)
   def options(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.OPTIONS, url)
-  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.valueOf(method), Left(url))
-  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(method, Left(url))
-  def httpRequest(method: HttpMethod, urlOrURI: Either[Expression[String], Uri]): HttpRequestBuilder =
+  def httpRequest(method: Expression[String], url: Expression[String]): HttpRequestBuilder = httpRequest(Left(method), Left(url))
```
slandelle, line 37: This is a binary breaking change. You cannot change existing signatures.

Honestly, I've tried. And failed.
If things were that simple, I would have done it myself.
I ended up with the Scala compiler no longer being able to pick my implicit conversions, causing tests to no longer compile.

slandelle (changes requested):

gatling-http/src/main/scala/io/gatling/http/request/builder/Http.scala
```diff
@@ -27,15 +27,15 @@ import io.netty.handler.codec.http.HttpMethod
...
   def patch(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.PATCH, url)
   def head(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.HEAD, url)
   def delete(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.DELETE, url)
   def options(url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.OPTIONS, url)
-  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(HttpMethod.valueOf(method), Left(url))
-  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(method, Left(url))
-  def httpRequest(method: HttpMethod, urlOrURI: Either[Expression[String], Uri]): HttpRequestBuilder =
+  def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(HttpMethod.valueOf(method)), Left(url))
+  def httpRequest(method: HttpMethod, url: Expression[String]): HttpRequestBuilder = httpRequest(Right(method), Left(url))
+  def httpRequest(method: Either[Expression[String], HttpMethod], urlOrURI: Either[Expression[String], Uri]): HttpRequestBuilder =
```
slandelle, line 39: Sorry, but I don't want to add any extra method, in particular one that would only serve 1 single person. On the contrary, I want to remove some methods, in particular the one that exposes `HttpMethod` (a netty type) and the one that uses `Uri` that's actually an internal.

Ultimately (in the next minor release), the only public methods should be:

```
def get(url: Expression[String]): HttpRequestBuilder = ???
def put(url: Expression[String]): HttpRequestBuilder = ???
def post(url: Expression[String]): HttpRequestBuilder = ???
def patch(url: Expression[String]): HttpRequestBuilder = ???
def head(url: Expression[String]): HttpRequestBuilder = ???
def delete(url: Expression[String]): HttpRequestBuilder = ???
def options(url: Expression[String]): HttpRequestBuilder = ???
// changed method from String to Expression[String]
def httpRequest(method: Expression[String], url: Expression[String]) = ???
```

The problem is that if we try to keep the existing `def httpRequest(method: String, url: Expression[String]): HttpRequestBuilder = ???` to maintain binary compatibility, some code no longer compiles because the compiler can't figure out the implicit conversions:

```
// no longer compiles
http("Request").httpRequest("MY_METHOD", "/")
```

I really think there's no solution and this change will have to wait.

aartajew: Yeah, cleaning it up sounds reasonable. When do you think this could be implemented?

slandelle: > When do you think this could be implemented?

Ni idea.
We have to wait for the next minor release.
A new minor release is not just a number bump for us. It requires some effort on the Enterprise side, so we only cut one when there's enough value, which is not the case at this time.

> let me know if you want me to kick start these changes to speed things up :)

This change is pretty simple once you afford to break binary compatibility. I already have a local branch with it.

What we need is new features with significant value. I can't think of any low hanging fruit.

Issue #4670, opened 2025-12-09
SSE: JSON data field is incorrectly serialized as escaped string instead of JSON object (contradicts documentation)


kyu4583:
### Description

When SSE events are converted to JSON for checks, the `data` field is always serialized as an escaped string, even when the original data is valid JSON. This contradicts the [official documentation](https://docs.gatling.io/reference/script/sse/#checks) which shows that JSON data should be preserved as a JSON structure.

### Documentation vs Actual Behavior

The documentation states that this SSE event:
```
id: 77535
event: snapshot
data: [{"symbol":"BTC","price":4628.27},{"symbol":"ETH","price":3249.19}]
```

should be turned into:
```json
{
  "id": 77535,
  "event": "snapshot",
  "data": [
    {"symbol":"BTC","price":4628.27},
    {"symbol":"ETH","price":3249.19}
  ]
}
```

However, the current implementation produces:
```json
{"event":"snapshot","id":"77535","data":"[{\"symbol\":\"BTC\",\"price\":4628.27},{\"symbol\":\"ETH\",\"price\":3249.19}]"}
```

The `data` field is a string with escaped quotes instead of a JSON array.

### Steps to Reproduce

1. Set up an SSE endpoint that sends events in the following format:
```
id: 77535
event: snapshot
data: [{"symbol":"BTC","price":4628.27},{"symbol":"ETH","price":3249.19}]
```

2. Create a Gatling simulation to capture the SSE message:
```java
public class SseTest extends Simulation {

    HttpProtocolBuilder httpProtocol = http
            .baseUrl("http://localhost:8080")
            .acceptHeader("text/event-stream");

    ScenarioBuilder scn = scenario("SSE Test")
            .exec(
                    sse("SSE Connect")
                            .get("/test-sse")
                            .await(30).on(
                                    sse.checkMessage("Receive message")
                                            .check(bodyString().saveAs("sseMessage"))
                            )
            )
            .exec(session -> {
                String message = session.getString("sseMessage");
                System.out.println("Received SSE message: " + message);
                return session;
            })
            .exec(sse("SSE Close").close());

    {
        setUp(scn.injectOpen(atOnceUsers(1))).protocols(httpProtocol);
    }
}
```

3. Run the simulation and observe the output:
```
Received SSE message: {"event":"snapshot","id":"77535","data":"[{\"symbol\":\"BTC\",\"price\":4628.27},{\"symbol\":\"ETH\",\"price\":3249.19}]"}
```

### Impact

[...]

PR #4672, opened 2025-12-10
fix: when serializing SSE events, convert numeric id to number, close #4671

### LINKED ISSUE #4671: SSE: numeric id field is serialized as string instead of number (contradicts documentation)
### Description

Following the fix in #4670, I noticed that the `id` field has a similar inconsistency with the documentation.

The [documentation example](https://docs.gatling.io/reference/script/sse/#checks) shows:
```json
{
  "id": 77535,
  "event": "snapshot",
  "data": [...]
}
```

But the current implementation produces:
```json
{
  "id": "77535",
  "event": "snapshot",
  "data": [...]
}
```

### Proposed Solution (PR opened)

Apply a similar pattern to the `data` field fix: detect if the id value is numeric, and if so, serialize it as a number.

Edge cases handled:
- Leading zeros (e.g., `"007"`) → kept as string (invalid JSON number)
- Single `"0"` → converted to number
- Non-numeric (e.g., `"abc123"`) → kept as string

I've opened a PR with this fix: #4672

kyu4583:
Motivation:
Similar to #4670, the `id` field should be serialized as a number when the value is numeric, matching the documentation example.

Changes:
- Detect numeric id values and serialize without quotes
- Handle edge cases: leading zeros kept as string, single "0" as number
- Add tests for numeric, non-numeric, leading zeros, and single zero cases

Note:
The numeric check uses `forall(_.isDigit)` which is a simple character comparison loop with no memory allocation. Given that SSE id values are typically short (a few digits), the overhead is negligible - similar to the existing `charAt` checks for JSON data detection.

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/action/sse/fsm/ServerSentEvent.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/action/sse/fsm/ServerSentEvent.scala
@@ -31,7 +31,20 @@ final case class ServerSentEvent(
       sb.append("\"event\":\"").append(value).append("\",")
     }
     id.foreach { value =>
-      sb.append("\"id\":\"").append(value).append("\",")
+      sb.append("\"id\":")
+      val length = value.length
+      if (
+        length > 0 &&
+          value.forall(_.isDigit) &&
+          (length == 1 || value.charAt(0) != '0')
+      ) {
+        // numeric id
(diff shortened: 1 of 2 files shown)


slandelle:
Declined for the reasons explains in here: https://github.com/gatling/gatling/issues/4671#issuecomment-3635886263

PR #4679, opened 2026-02-02
fix: SseStreamDecoder handle multiline data lines correctly, close #4678

### LINKED ISSUE #4678: SSE: SseStreamDecoder should concatenate multiple data: lines (SSE compliance)
### Description

SSE events may contain multiple `data:` lines. In such cases, the event `data` should be treated as a single value where each `data:` line is concatenated with a newline character (`\n`) between lines.

Spec reference:
- HTML Standard (WHATWG) — Server-sent events: for each `data` field line, the user agent appends the field value to the `data` buffer and then appends a single LF (`\n`), which results in multiple `data:` lines being concatenated with newlines (the trailing newline is removed when dispatching the event).
  https://html.spec.whatwg.org/multipage/server-sent-events.html

However, the current `SseStreamDecoder` effectively keeps only the last `data:` line of an event, so multiline SSE payloads are decoded incorrectly.

### Reproduction

Input stream:

```text
data: hello
data: world
```

Expected:
- One `ServerSentEvent` with `data = Some("hello\nworld")`.

Actual:
- One `ServerSentEvent` with `data = Some("world")` (previous `data:` lines are lost).

This also affects multiline payload patterns (e.g., multiline JSON/text), and cases where other fields (`event:`, `id:`, `retry:`) are interleaved with multiple `data:` lines.

### Proposed Solution (PR opened)

Accumulate multiple `data:` lines and join them with `\n` when emitting `ServerSentEvent` upon the blank line terminator.

To minimize overhead in the common single-line case, keep the first `data:` line as a `String` and promote to a `StringBuilder` only when a second `data:` line is encountered.

PR: #4679

kyu4583:
### Motivation:
`SseStreamDecoder` currently keeps only the last `data:` line of an event, which causes multiline SSE payloads to be decoded incorrectly.

### Changes:
- Concatenate multiple `data:` lines with `\n` when flushing an event
- Preserve empty `data:` lines in multiline events (e.g., `data:`)
- Minimize overhead for the common single-line case: keep the first line as `String` and promote to `StringBuilder` only when a second `data:` line is encountered
- Add tests covering:
  - Multiline JSON-like payload
  - Plain text multiline payload
  - Interleaved fields (`event`, `id`, `retry`) with multiple `data:` lines
  - Single `data:` line behavior
  - Empty `data:` lines within multiline payload

### Before/After Example

SSE data (multiline JSON) sent by server:
```text
event: json
id: 1
retry: 1000
data: {
data:   "foo": "bar",
data:   "count": 123
data: }
```

Gatling check snippet (trimmed):
```java
sse.checkMessage("json-data")
  .matching(jsonPath("$.event").is("json"))
  .check(jsonPath("$.data.foo").is("bar"))
  .check(bodyString().transform(v -> {
      System.out.println("[SSE Message] json-data received: " + v);
      return v;
  }));
```

**Before:**
```text
[SSE Message] json-data received: {"event":"json","id":"1","data":"}","retry":1000}
```

**After:**
```text
[SSE Message] json-data received: {"event":"json","id":"1","data":{
  "foo": "bar",
  "count": 123
},"retry":1000}
```

Fixes #4678

### DIFF
--- a/gatling-http/src/main/scala/io/gatling/http/action/sse/fsm/SseStreamDecoder.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/action/sse/fsm/SseStreamDecoder.scala
@@ -37,7 +37,8 @@ final class SseStreamDecoder extends Utf8ByteBufCharsetDecoder {
   import SseStreamDecoder._
 
   private[this] var pendingEvent: Option[String] = None
-  private[this] var pendingData: Option[String] = None
+  private[this] var pendingData: String = null
+  private[this] var pendingDataBuilder: StringBuilder = null
   private[this] var pendingId: Option[String] = None
   private[this] var pendingRetry: Option[Int] = None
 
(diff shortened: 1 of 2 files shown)


kyu4583:
> Hello,
> 
> Thanks a lot for contributing! Just one small nit.
> 
> Could you also please our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw)?
> 
> Cheers!

Done. Thank you!

Issue #4471, opened 2023-09-26
HTTP: Make WsLogger use the same slf4j logger as HTTP (io.gatling.http.engine.response)


slandelle:
Way too cumbersome to have to enable and sync 2 different loggers.

slandelle:
@Amerousful FYI

Amerousful:
Roger that

Amerousful:
@slandelle Hi. You know, now I don't think this change was a good idea. In my opinion, it's better to have the option to filter your loggers, meaning you can enable or disable them as needed. 

I comprehend that WebSocket operates on the HTTP protocol, yet it is a distinct protocol.
Also, I found as an example - SSE (Server Sent Event) which has a separated logger.
```xml
<logger name="io.gatling.http.action.sse.fsm" level="DEBUG" />
```
https://gatling.io/docs/gatling/reference/current/http/sse/#debugging

Could I ask you to reconsider this change?

Amerousful:
@slandelle Hi again! Would you be able to review the message above?

PR #4705, opened 2026-08-07
fix: categorize WebSocket traffic logs like HTTP (DEBUG failures, TRACE all)

### LINKED ISSUE #4468: HTTP: Whole traffic from WebSocket is logged as a DEBUG - can't filter failed messages
Hello,

I'm using Gatling 3.9.5 and found that issue reported here: https://github.com/gatling/gatling/issues/4439 is still visible. Original ticket was not reopened so I raised another one

During simulation with WebSocket every log related to WebSocket traffic is using DEBUG level. It means even if I try to log to file only failed messages Gatling will store everything. The most annoying issue of that is a log file with a couple of GBs which is harder to store/parse after tests

Is it possible to fix this and use logic similar to HTTP where using DEBUG for logging only failed messages will be logged, and TRACE for everything?

My logback.xml content:
```
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
        <encoder>
            <pattern>%d{HH:mm:ss.SSS} [%-5level] %logger{15} - %msg%n%rEx</pattern>
        </encoder>
        <immediateFlush>false</immediateFlush>
    </appender>
    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
        <file>target/gatling/LastRun.log</file>
        <append>true</append>
        <encoder>
            <pattern>%d{HH:mm:ss.SSS} [%-5level] %logger{15} - %msg%n%rEx</pattern>
        </encoder>
    </appender>
    <!-- uncomment and set to DEBUG to log all failing HTTP requests -->
    <!-- uncomment and set to TRACE to log all HTTP requests -->
    <logger name="io.gatling.http.engine.response" level="DEBUG"/>
    <logger name="io.gatling.http.action.ws.fsm" level="DEBUG" />
    <root level="WARN">
        <appender-ref ref="FILE"/>
<!--        <appender-ref ref="CONSOLE"/>-->
    </root>
</configuration>
```

BR
Mateusz

arimu1:
### Motivation

Fixes #4468.

WebSocket FSM path logged almost all traffic-related events at `DEBUG`, so enabling WebSocket logging (or the package logger) dumped entire successful traffic into multi-GB files. HTTP already uses `DEBUG` for failures only and `TRACE` for everything via `io.gatling.http.engine.response`.

This aligns WebSocket with that model, following the approach agreed on the issue:

- outbound: TRACE at send; retain for later dump when a check is present
- inbound matching a check: TRACE dump on success, DEBUG dump on failure (`WsLogger`)
- inbound non-matching / wrong type: TRACE only (keep buffered for a later failed-check dump)
- high-volume FSM lifecycle (timeout schedule/cancel): TRACE

Also fixes a buffer-clear bug: `logCheck(OK)` was called before applying checks and on non-matching frames, which cleared the inbound buffer so later KO dumps (timeout / failed check) often lost received messages.

### Modifications

- `WsIdleState`: send frame logs → TRACE; `logOk` only when there is no check; with checks, defer dump until check outcome
- `WsPerformingCheckState`: matching/non-matching/unmatched traffic → TRACE; call `logCheck(OK)` only after check success; do not `logCheck` on non-matching (retain buffer)
- `WsFsm`: timeout schedule/cancel → TRACE; timeout trigger stays DEBUG
- `logback.dummy`: document that `io.gatling.http.engine.response` covers HTTP **and** WebSocket; FSM package at TRACE for state-machine noise

`WsLogger` already uses `HttpTracing.LOGGER` (`io.gatling.http.engine.response`) from #4471.

### Result

With:

```xml
<logger name="io.gatling.http.engine.response" level="DEBUG"/>
```

only failed WebSocket checks (and failed HTTP requests) are logged.

With `TRACE` on the same logger, full traffic dumps are logged.

FSM package logger no longer floods DEBUG with full message bodies on successful traffic.

### Testing

- `sbt gatling-http/compile` (Temurin 21) — success
- No open prior PR for #4468 (searched issue number + WebSocket/WsLogger log keywords)

### DIFF
--- a/gatling-core/src/main/resources/logback.dummy
+++ b/gatling-core/src/main/resources/logback.dummy
@@ -8,12 +8,12 @@
 		<immediateFlush>false</immediateFlush>
 	</appender>
 
-	<!-- uncomment and set to DEBUG to log all failing HTTP requests -->
-	<!-- uncomment and set to TRACE to log all HTTP requests -->
+	<!-- uncomment and set to DEBUG to log failing HTTP and WebSocket requests -->
+	<!-- uncomment and set to TRACE to log all HTTP and WebSocket requests -->
 	<!--<logger name="io.gatling.http.engine.response" level="TRACE" />-->
 
-	<!-- uncomment to log WebSocket events -->
-    <!--<logger name="io.gatling.http.action.ws.fsm" level="DEBUG" />-->
+	<!-- uncomment and set to TRACE to log WebSocket FSM events (connect/send/receive state machine) -->
+    <!--<logger name="io.gatling.http.action.ws.fsm" level="TRACE" />-->
 
 	<!-- uncomment to log SSE events -->
     <!--<logger name="io.gatling.http.action.sse.fsm" level="DEBUG" />-->
--- a/gatling-http/src/main/scala/io/gatling/http/action/ws/fsm/WsFsm.scala
+++ b/gatling-http/src/main/scala/io/gatling/http/action/ws/fsm/WsFsm.scala
@@ -55,6 +55,7 @@ final class WsFsm(
   private[fsm] def scheduleTimeout(dur: FiniteDuration): Unit = {
     currentTimeout = eventLoop.schedule(
       () => {
+        // timeout fired (often a failed check) — keep at DEBUG for troubleshooting
         logger.debug(s"Timeout ${currentTimeout.hashCode} triggered")
         currentTimeout = null
         execute(currentState.onTimeout())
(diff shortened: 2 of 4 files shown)


slandelle:
Hi,

Thanks for your contribution. I'm currently off and won't be able to review until I'm back. Please be patient :)

In the meantime, could you please sign our [CLA](https://docs.google.com/forms/d/1hbpxVqJ5hIYYJuOjOyOzub9fJFGFGvvfnB5w5LCWdUw)?

arimu1:
Thanks — I've signed the Gatling CLA via the Google form. Happy to wait until you're back from leave for the review.

slandelle (approved):

slandelle:
I had to cherry-pick due to our internal workflow, but I had no request for change.
Thanks a lot @arimu1 !

